PR #28, opened 2014-08-04
msvc: allow dlls without lib file

bernhard-b:
On Windows it is possible to create a dll consisting only of a resource (message) file which is used for the eventlog. Because it doesn't export any function the msvc linker doesn't create a lib file. Due to that boost build always thinks that the dll needs to be build. i have registered the generator twice with the feature suppress-import-lib (there is no use/doc of it) and only generate the IMPORT_LIB target for one. i haven't figured out another way to achieve that.
What is missing that /NOENTRY needs to be passed to the linker. The pull request is to gain ideas/comments how to fully implement the feature for dlls without any exports.
regards,
bernhard


swatanabe:
This appears to have been merged, but was never closed.

pdimov:
FWIW, as mentioned, using `/NOENTRY` breaks building python extensions such as the MPI one, because they use `<suppress-import-lib>true` because they don't need to export anything, but are not resource DLLs. See https://github.com/boostorg/build/issues/734.

Issue #54, opened 2015-01-22
Type Checking


frenchtoast747:
I'm continuously working on and using the Python port and I've been bitten by a similar bug several times: I pass in a string when the function is expecting a list. The function will usually continue on as normal and later on show some cryptic error message about how it can't find some target named '_', 'L', or any other various first letters of one of my targets.

Some of the core boost-build functions and methods already contain asserts for type checking, but there are a lot that don't. I've developed somewhat of a solution that utilizes `bjam_signature`. It uses the signature passed in to determine the appropriate type of a parameter and raises a `TypeError` if, say, a string is passed in where a list was expected.

I've already tried implementing this solution and it works. However, there are some functions that require a list as a parameter and account for parameters that might be passed a string. For example, `feature.compose`:

```
# feature.py
@bjam_signature((["composite_property_s"], ["component_properties_s", "*"]))
def compose (composite_property_s, component_properties_s):
    # ...
    component_properties_s = to_seq (component_properties_s)
```

And in `type.register`, `feature.compose` is called with a string instead of a list:
```
if base_type:
        feature.compose ('<target-type>' + type, replace_grist (base_type, '<base-target-type>'))
        feature.compose ('<base-target-type>' + type, '<base-target-type>' + base_type)
```

My solution doesn't work when variables are allowed to be either a list or a string even though the bjam signature says it should be a list. My question: would it be permissible to enforce this strict(er) typing by implementing the `bjam_signature` type check and removing all type coercions in the core? Or should I try the next idea:

Instead of or along with utilizing the `bjam_signature` type check, it would be nice to place type checking asserts for each function/method that cannot support a `bjam_signature`.

Both solutions would require a small change to the engine code to enable the the `PYTHONOPTMIZE` flag by default (so all asserts and `if __debug__` blocks are removed) and allow the optimization flag to be disabled by another flag for development/debugging purposes.

vprus:
Aaron,

personally I think that stricter checking would be fine. E.g. passing a list as the second parameter to feature.compose seems quite reasonable to me.

frenchtoast747:
Thanks Vladimir! I'll implement my solution and create a pull request soon.

vprus:
Since the pull request was merged quite some time ago, I assume the issue is also resolved.

frenchtoast747:
Yes it is. Thanks!

PR #55, opened 2015-01-31
Develop mode type checking

frenchtoast747:
These changes change the core engine by loading the python modules in optimization mode by default and allow specifying a flag to disable optimized mode. When optimized mode is on, all `assert` statements as well as `if __debug__` statements are completely removed from the AST before being compiled to bytecode. Both are being used within the `bjam_signature` decorator and a majority of the code within the `build` package.

I am still in the process of porting our large library written in Jam over to Python. I do my best to try and determine the signature of a function/rule based on the usage of a parameter on the Jam side, however, I've been bitten by the same  bug where I've passed in a string when it has expected a list.

After completing the type checking, I've found several other type-bugs which has helped me solve a lot of problems I've been having.

This addresses #54.

### DIFF
--- a/src/build/alias.py
+++ b/src/build/alias.py
@@ -29,7 +29,7 @@ import targets
 import property_set
 from b2.manager import get_manager
 
-from b2.util import metatarget
+from b2.util import metatarget, is_iterable_typed
 
 class AliasTarget(targets.BasicTarget):
 
@@ -37,9 +37,17 @@ class AliasTarget(targets.BasicTarget):
         targets.BasicTarget.__init__(self, *args)
 
     def construct(self, name, source_targets, properties):
+        if __debug__:
+            from .virtual_target import VirtualTarget
+            assert isinstance(name, basestring)
+            assert is_iterable_typed(source_targets, VirtualTarget)
+            assert isinstance(properties, property_set.PropertySet)
         return [property_set.empty(), source_targets]
 
     def compute_usage_requirements(self, subvariant):
+        if __debug__:
+            from .virtual_target import Subvariant
+            assert isinstance(subvariant, Subvariant)
         base = targets.BasicTarget.compute_usage_requirements(self, subvariant)
         # Add source's usage requirement. If we don't do this, "alias" does not
         # look like 100% alias.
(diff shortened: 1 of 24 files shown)


frenchtoast747:
I've started to have second thoughts about the various type checking `assert` statements.

Rather than having:
```
assert is_iterable_typed(property_set, basestring)
```

It should probably be something like:
```
assert is_property_set_literal(properties)
```

Or even:
```
assert is_propert_set(properties)
```

So that if the majority of the python infrastructure is changed to be passing instances of property sets instead of the string literals, only the assert function will need to be changed.

vprus:
Aaron,

sorry it took me a while to get to this. Do you want me to review this version, or there's a better one elsewhere?

frenchtoast747:
This is all I have at the moment. I'd appreciate it if you could review this version. Thanks!

frenchtoast747:
Also, no worries! I've been using it on a personal branch which has been working just fine. Let me know if I should rebase this on top of the `develop` branch.

frenchtoast747:
I have used `__debug__` when a specific type doesn't exist in a module and cannot be imported at the top of the module due to circular dependencies and when that type doesn't need to be imported during normal operation of that function. For example, there are quite a few instances of that where the `feature` module needs access to the `PropertySet` class. 

The last usage is within `util` specifically for the `bjam_signature` function since `bjam_signature` is needed for normal operation, but the type checking is not. Any of the helper functions used by the `__debug__` section in `bjam_signature` have also been guarded by a `__debug__`.

I've left some of the helper functions outside of the `__debug__`, such as `is_iterable` and `is_iterable_typed`, so that they can be imported at the top of the modules and will not throw an error when run in optimized mode.

vprus:
Aaron, if you rebase this change I'll be happy to apply it.

frenchtoast747:
I have squash-rebased the typechecking branch on develop.

Just curious, is there a way to run the tests against the Python port? I've been using my in house testing for that.

vprus:
I've cherry-picked this change, thanks!

If you're on Linux, testing Python port is basically `python test_all.py --python`, it's fairly healthy I believe. On Windows, most tests fail due to added "threading=multi" in the target paths, not yet sure why.

frenchtoast747:
I have written tests that run our codebase with both the Jam and the Python versions to ensure that our ported Python produces the same output as the existing Jam and I've noticed the difference with "threading-multi". If I ever figure out the cause, I'll be sure to submit a patch.

PR #60, opened 2015-02-26
Add support for more standard library options within sun.jam

jzmaddock:
The latest Oracle compiler has several std lib options, all of which work with Boost to some degree or another:

* default C++03 - Rogue Wave
* stlport C++ 03
* Apache C++03
* C++11 - a version of GNU Libstdc++

This patch provides a means to switch between them.

### DIFF
--- a/src/tools/sun.jam
+++ b/src/tools/sun.jam
@@ -22,6 +22,14 @@ feature.extend stdlib : sun-stlport ;
 feature.compose <stdlib>sun-stlport
     : <cxxflags>-library=stlport4 <linkflags>-library=stlport4
     ;
+feature.extend stdlib : apache ;
+feature.compose <stdlib>apache
+    : <cxxflags>-library=stdcxx4 <linkflags>-library=stdcxx4
+    ;
+feature.extend stdlib : cxx11 ;
+feature.compose <stdlib>cxx11
+    : <cxxflags>-std=c++11 <linkflags>-std=c++11
+    ;
 
 rule init ( version ? : command * : options * ) 
 {
--- a/src/tools/sun.jam
+++ b/src/tools/sun.jam
@@ -23,8 +23,7 @@ generators.override sun.searched-lib-generator : searched-lib-generator ;
 #    also available via -std=sun03.
 # 2) C++03 mode + STLport, selected via the -library option.
 # 3) C++03 mode plus the Apache std lib, selected via the -library option.
-# 4) C++03 mode in g++ compatibility mode, and GNU libstdc++3, selected via -std=c++03.
-# 5) C++11 mode in g++ compatibility mode, and GNU libstdc++3, selected via -std=c++11.
+# 4) C++03 or c++11 in g++ compatibility mode, and GNU libstdc++3, selected via -std=c++03.
 #
 # Note that the -std, -library and -compat compiler switches appear to be largely mutually
 # incompatible, and that going forward the -std switch seems to be the prefered one.
@@ -39,13 +38,9 @@ feature.extend stdlib : apache ;
 feature.compose <stdlib>apache
     : <cxxflags>-library=stdcxx4 <linkflags>-library=stdcxx4
     ;
-feature.extend stdlib : cxx03 ;
-feature.compose <stdlib>cxx03
+feature.extend stdlib : gnu ;
+feature.compose <stdlib>gnu
     : <cxxflags>-std=c++03 <linkflags>-std=c++03
-    ;
-feature.extend stdlib : cxx11 ;
-feature.compose <stdlib>cxx11
-    : <cxxflags>-std=c++11 <linkflags>-std=c++11
     ;
 
 rule init ( version ? : command * : options * ) 
--- a/src/tools/sun.jam
+++ b/src/tools/sun.jam
@@ -17,7 +17,20 @@ toolset.inherit  sun : unix ;
 generators.override sun.prebuilt : builtin.lib-generator ;
 generators.override sun.prebuilt : builtin.prebuilt ;
 generators.override sun.searched-lib-generator : searched-lib-generator ;
-
+#
+# There are no less than 5 standard library options:
+# 1) The default, which uses an old version of the Rogue Wave std lib,
+#    also available via -std=sun03.
+# 2) C++03 mode + STLport, selected via the -library option.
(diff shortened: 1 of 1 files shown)


swatanabe:
AMDG

One minor issue I have with this is the naming.
The existing option is called <stdlib>sun-stlport.
I believe this is to avoid a conflict with
stlport.jam.  I know that Boost.Build doesn't
have direct support for stdcxx, right now,
but do you know if the apache library can
be used as a replacement standard library
in general, so that this option could
theoretically conflict with a hypothetical
apache.jam?

The other issue is that it seems somewhat
non-intuitive that the standard library
option should set the language version.
Is there any way around this?  Does -std=c++03
have any other side effects?  I know with
gcc, the default is actually -std=gnu++NN,
rather than -std=c++NN, for example.

In Christ,
Steven Watanabe

jzmaddock:
On 03/03/2015 09:05, swatanabe wrote:
> AMDG
>
> One minor issue I have with this is the naming.
> The existing option is called <stdlib>sun-stlport.
> I believe this is to avoid a conflict with
> stlport.jam. I know that Boost.Build doesn't
> have direct support for stdcxx, right now,
> but do you know if the apache library can
> be used as a replacement standard library
> in general, so that this option could
> theoretically conflict with a hypothetical
> apache.jam?

I guess so, the apache library is the successor to the Rogue Wave 
library, but open source (http://stdcxx.apache.org/).

However, the library development appears to be dead, and I've never 
heard of it being used with any compiler other than Oracle's.

>
> The other issue is that it seems somewhat
> non-intuitive that the standard library
> option should set the language version.
> Is there any way around this? Does -std=c++03
> have any other side effects? I know with
> gcc, the default is actually -std=gnu++NN,
> rather than -std=c++NN, for example.

It would be more accurate to call these binary-compatibility options 
rather than library options, but there is one standard library for each 
binary option: you absolutely cannot mix and match.  There are also no 
gnu++ options.  To summarise we have:

C++03, sun object format, sun lib:  default.
C++03, sun object format, STLPort, -library=-stlport4
C++03, sun object format, Apache, -library=-stdcxx4
C++03, GNU binary, GNU lib, -std=c++03
C++11, GNU binary, GNU lib, -std=c++11

There is also -std=sun-03 which is the same as one of the first 3 (not 
sure which).  And c++11 mode only works in GNU mode with GNU libstdc++3.

The above 5 options are mutually incompatible, and -compat=, -library= 
and -std= options cannot be mixed and matched - in short their command 
line syntax is a bit of mess :-(

John.

swatanabe:
AMDG

On 03/03/2015 03:54 AM, jzmaddock wrote:
> I guess so, the apache library is the successor to the Rogue Wave 
> library, but open source (http://stdcxx.apache.org/).
> 
> However, the library development appears to be dead, and I've never 
> heard of it being used with any compiler other than Oracle's.
> 

All right.  In that case there shouldn't be a problem.

> It would be more accurate to call these binary-compatibility options 
> rather than library options, but there is one standard library for each 
> binary option: you absolutely cannot mix and match.  There are also no 
> gnu++ options.  To summarise we have:
> 
> C++03, sun object format, sun lib:  default.
> C++03, sun object format, STLPort, -library=-stlport4
> C++03, sun object format, Apache, -library=-stdcxx4
> C++03, GNU binary, GNU lib, -std=c++03
> C++11, GNU binary, GNU lib, -std=c++11
> 
> There is also -std=sun-03 which is the same as one of the first 3 (not 
> sure which).  And c++11 mode only works in GNU mode with GNU libstdc++3.
> 
> The above 5 options are mutually incompatible, and -compat=, -library= 
> and -std= options cannot be mixed and matched - in short their command 
> line syntax is a bit of mess :-(
> 

One more question.  If you provide *both* -std=c++03
and -std=c++11, is it a hard error, does it do something
strange, or does it pick one in a predictable way?
I'd rather have -std=c++03 and -std=c++11 use the same
<stdlib> property and have some other way to distinguish
them.  Hmmm.  There was talk a while back about a feature
to control c++11/c++03.  Let me check on that.

In Christ,
Steven Watanabe

jzmaddock:
One more question. If you provide *both* -std=c++03
> and -std=c++11, is it a hard error, does it do something
> strange, or does it pick one in a predictable way?
> I'd rather have -std=c++03 and -std=c++11 use the same
> <stdlib> property and have some other way to distinguish
> them. Hmmm. There was talk a while back about a feature
> to control c++11/c++03. Let me check on that.
>
It works, and it picks the rightmost one on the command line.

However you would have to be very careful not to break the 
stlport/apache/Cstd library options as you cannot mix the -std= command 
line option with either of the -library= or -compat= options (it's a 
hard error and bails right out).

swatanabe:
AMDG

On 03/03/2015 10:07 AM, jzmaddock wrote:
> One more question. If you provide *both* -std=c++03
>> and -std=c++11, is it a hard error, does it do something
>> strange, or does it pick one in a predictable way?
>> I'd rather have -std=c++03 and -std=c++11 use the same
>> <stdlib> property and have some other way to distinguish
>> them.
>>
> It works, and it picks the rightmost one on the command line.
> 
> However you would have to be very careful not to break the 
> stlport/apache/Cstd library options as you cannot mix the -std= command 
> line option with either of the -library= or -compat= options (it's a 
> hard error and bails right out).
> 

Okay, I propose that the stdlib feature
should have a single value for libstdc++
which should set -std=c++03.

The result will be:

<stdlib>native -> no options
<stdlib>native + <cxxflags>-std=c++11 -> -std=c++11
<stdlib>sun-stlport|apache -> -libary=...
<stdlib>sun-stlport|apache + <cxxflags>-std=c++11 -> compiler error
<stdlib>libstdc++ -> -std=c++03
<stdlib>libstdc++ + <cxxflags>-std=c++11 -> -std=c++03 -std=c++11
  -> -std=c++11

which I think is sensible.  (Note: we don't have
to do anything special to make sure that -std=c++11
appears after -std=c++03.  It matches the default
lexicographical sort order.)

In Christ,
Steven Watanabe

jzmaddock:
OK, I've updated and tested the patch accordingly.  Now uses stdlib=gnu to select libstdc++3.

Aside: I still think we need a standard conformance selection feature, it's a pain to use cxxflags as that doesn't work if you want to test multiple configurations.

jzmaddock:
Thanks!

PR #61, opened 2015-03-08
Create a simple interactive shell for evaluating Jam.

frenchtoast747:
One of the biggest complaints I get about using Boost.Build is that Jam is confusing. There isn't any sort of a sandbox to play around in with the language itself. The best solution I've come up with is throwing all the  statements I want to eval at the beginning of my Jamroot immediately followed by an `exit ;`.

One of the best resources at my disposal while learning Python was the interactive shell. So, I've created something similar to automating the above solution. In the long run, I'd like to actually use the internal parser, but it's so tightly integrated with the compile phase that I just opted with this stupid simple solution for now.

To use, either directly run `interpreter.py` (`-e` to specify where the `b2` executable is located if it's not in your path) or run `b2 --python --interactive` to bring up the shell. Then, issue simple Jam statements.

An example session
```
Boost.Build Interactive Shell
Type "help ;" for a simple tutorial.
________________________________________________________________________________
>>> # simple variable declaration
>>> x = asdf fdsa qwerty ;
>>> # simple variable evaluation;
>>> # the user doesn't have to manually "echo" everything
>>> $(x) ;
asdf fdsa qwerty
>>> y = 1 2 3 ;
>>> # expected expansions
>>> $(x)::$(y) ;
asdf::1 asdf::2 asdf::3 fdsa::1 fdsa::2 fdsa::3 qwerty::1 qwerty::2 qwerty::3
>>> # rule declarations
>>> rule say-hello ( name ? )
... {
...     name ?= there ;
...     echo Hello, $(name). ;
... }
>>> # the rules are remembered and can be invoked
>>> say-hello Aaron ;
Hello, Aaron.
>>> x = prefix/$(x) ;
>>> # it's possible to import another module on the BOOST_BUILD_PATH
>>> import regex ;
>>> # rules can be evaluated without an echo too
>>> [ regex.replace-list $(x) : prefix/ : new-prefix/ ] ;
new-prefix/asdf new-prefix/fdsa new-prefix/qwerty
```

### DIFF
--- a/src/build_system.py
+++ b/src/build_system.py
@@ -454,6 +454,12 @@ def main_real():
 
     import b2.build.configure as configure
 
+    if '--interactive' in sys.argv:
+        from interpreter import InteractiveShell
+        exe = sys.argv[0]
+        boost_build_path = bjam.variable('BOOST_BUILD_PATH')
+        return InteractiveShell(exe, boost_build_path).interact()
+
     if "--version" in sys.argv:
         from b2.build import version
         version.report()
--- a/src/interpreter.py
+++ b/src/interpreter.py
@@ -1,0 +1,343 @@
+import re
+import os
+import tempfile
+import subprocess
+import shutil
+import sys
+
+
+RULE_DECL_REGEX = re.compile(r'\s*rule\s+(?P<name>[\w\d_-]+)(\s+\( .*? \))?')
+MODULE_DECL_REGEX = re.compile(r'\s*module\s+(?P<name>[\w\d_-]+)\s+')
+VARIABLE_DECL_REGEX = re.compile(r'\s*(local\s+[\w\d_-]+)|([\w\d_-]+\s+\=)')
+IMPORT_DECL_REGEX = re.compile(r'\s*import .*?;')
+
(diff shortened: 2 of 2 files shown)


swatanabe:
AMDG

  This is a good idea, in theory.  In the
long term, I want to implement a debugger
as part of b2.  The main issue I have is
that it's basically impossible to make the
system of saving statements and re-executing
them handle all situations.  What you really
need to do is to compile and run each statement
in a single module.  This means adding a new
builtin rule, which would be implemented like:

- Set up the lexer to read from the string
- call the parser
- Strip off local from the top level node
- wrap the AST in a call to echo
- call function_compile
- call function_run

In Christ,
Steven Watanabe

frenchtoast747:
Agreed. In that case, I'll close this pull request so that the correct solution can be implemented. Thanks!

KayEss:
Does this not help people in the meantime, or is the concern that doing this will block out a better approach?

frenchtoast747:
In theory, after the better solution is implemented, there should only need to be two changes. There will be a need for more initialization code and a switch from calling the subprocess to `b2` to just calling the new builtin rule.

PR #63, opened 2015-03-11
Typo in the document for setting toolset option

yangcha:


jhunold:
Thanks for the patch.

PR #67, opened 2015-04-10
Typo in overview.xml

teeks99:
Typo, it should be 'Do not execute' not 'Do no execute'.


teeks99:
The utils get linked into core, not the other way around.

vprus:
Tom,

I've applied both patches, thanks!

PR #70, opened 2015-04-23
support aix-soname=svr4 option, with gcc for now

haubi:
On AIX, using Import Files it is possible to provide filename-based shared library versioning similar to SVR4/ELF platforms, as outlined in
https://gcc.gnu.org/install/configure.html#WithAixSoname

While this patch already works (with gcc) so far, I'd like to improve it a little but fail with jam: In rule link.dll.aix I want to build the list of local object files ($sources of type OBJ) to extract the public symbols from. Currently I do this in the actions link.dll.aix in the shell code - works, but feels ugly.
Hints welcome, thanks!

### DIFF
--- a/src/tools/builtin.jam
+++ b/src/tools/builtin.jam
@@ -226,6 +226,9 @@ feature.feature def-file : : free dependency ;
 
 feature.feature suppress-import-lib : false true : incidental ;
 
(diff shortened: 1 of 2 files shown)


eldiener:
All PRs need to be made against the 'develop' branch and not the 'master' branch, as you have done.

PR #75, opened 2015-04-29
Fix implicit-dependency scanner bug.

frenchtoast747:
There was a typo in the port for the scanners which caused generated headers to drop off.

See this email for more details:
http://lists.boost.org/boost-build/2015/04/27979.php

Thanks to Steven for the fix!


vprus:
Merged, thanks.

PR #78, opened 2015-05-05
use -undefined dynamic_lookup on darwin in python_for_extensions

### LINKED ISSUE #69: python_for_extensions should use -undefined dynamic_lookup on darwin
build/src/tools/python.jam says:

>    # Unlike most *nix systems, Mac OS X's linker does not permit undefined
    # symbols when linking a shared library. So, we still need to link against
    # the Python framework, even when building extensions.

Passing `-undefined dynamic_lookup` to the linker (instead of -lpython or -framework Python) permits undefined symbols and achieves the desired behavior. This has existed [since OS X 10.1](http://www.cmake.org/pipermail/cmake/2005-March/006249.html) so there should be no compatibility concerns.

Linking a module against one Python framework and importing it from another (ABI-compatible) interpreter causes the interpreter to segfault, so the current behavior is a pain point for e.g. Homebrew.

tdsmith:
Passing -undefined dynamic_lookup to the linker (instead of -lpython or
-framework Python) permits undefined symbols in shared libraries on OS
X. This allows a module to be linked against one Python framework and
imported from another.

Tested by building ledger, which uses Boost.Python, against a version of Boost built with and without this patch. In the "before" gist, libboost_python-mt.dylib has a linkage to `/usr/local/Frameworks/Python.framework/Versions/2.7/Python`; in the "after" gist, it does not, and the ledger python demo still runs correctly. https://gist.github.com/tdsmith/558243ad98ad288a16e0

Closes #69.

### DIFF
--- a/src/tools/python.jam
+++ b/src/tools/python.jam
@@ -962,15 +962,17 @@ local rule configure ( version ? : cmd-or-prefix ? : includes * : libraries ? :
     # symbols. If we linked to libpython, we would get duplicate symbols. So
     # declare two targets -- one for building extensions and another for
     # embedding.
-    #
-    # Unlike most *nix systems, Mac OS X's linker does not permit undefined
-    # symbols when linking a shared library. So, we still need to link against
-    # the Python framework, even when building extensions. Note that framework
-    # builds of Python always use shared libraries, so we do not need to worry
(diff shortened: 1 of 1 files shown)


vprus:
Thanks, cherry-picked.

PR #82, opened 2015-06-02
Xlcpp toolset build

camorton2:
creating a new toolset 'xlcpp' for new IBM compilers based on clang, leaving old toolset 'vacpp' for use with previous IBM compilers not based on clang


camorton2:
We use the clang front end, but its still an 'xlC' compiler

camorton2:
I decided to setup a new toolset mainly because we don't support all of clang's features so I might have to change some things as we investigate test failures.

src/tools/xlcpp.jam
```diff
@@ -0,0 +1,150 @@
+# Copyright Vladimir Prus 2004.
+# Copyright Toon Knapen 2004.
```
grafikrobot: You should update the copyrights to reflect the true authors, and dates.

camorton2: Can I just remove those lines?   I notice clang.hpp has no copyright other than

# Distributed under the Boost Software License, Version 1.0.
# (See accompanying file LICENSE_1_0.txt
# or copy at http://www.boost.org/LICENSE_1_0.txt)

I just copied and updated from vacpp.hpp and I have no idea who the original author is for that file.

grafikrobot: It needs to have some for of copyright.. To keep lawyers happy. If it's a mostly copied file, just keep the original copyrights and add a line for yourself to reflect your contribution.

camorton2:
Is there any further status on this request?  We would like to start using the new toolset for regular runs.

eldiener:
Is it possible to test this toolset out on Linux ? I applied the PR on Boost build on Windows but I could see from your changes it is not run there. But after running booststrap.bat to sync up Boost build versions,  I was able to run Boost build with my normal Windows toolsets ( gcc, clang, and vc++ ) without any problems. But to really test that this toolset actually works on Linux with the IBM compiler using clang it would be best if I could test it with the appropriate compiler. Is there a developer ( free ) version of the compiler I can test with on the IBM website ?

camorton2:
I tested it on linux with xlC

eldiener:
My point is that, while you can test it, if the developers of Boost libraries cannot test the compiler none of the Boost libraries are going to pay any attention to the compiler. If IBM expects Boost library developers to pay for a license of xlc I think they are in for a rude awakening. And without being able to test the compiler I highly doubt that any Boost library developer is going to approve of a library change because of xlc, whether you say that the change is necessary or not. Of course I could be wrong about other Boost developers and we do try to support compilers which are very difficult to be able to test with.

I talked online to someone from IBM about getting a free license for Boost developers to test the compiler on Linux. The answer I received is that while the person i was talking to online dealt with licenses he/she did not deal with any free licenses, if they exist. It was suggested to by the person I was chatting with online that I contact you as an IBM developer. If you are not an employee of IBM I do apologize.

camorton2:
In the recent past we have done regular testing and published the results on the trunk summary and we would like to start publishing our test results again with regular runs.  We have successfully submitted changes in the past (there's already a vacpp toolset), so I was hoping we could at least try rather than having to ship a separate patch when changes are required.
Changes in the past have been submitted by Chris Cambly or Michael Wong.

eldiener:
What do you mean by "submitted changes in the past" ? Do you mean you have created patches or pull requests against Boost libraries ?

camorton2:
We have done both.   Traditionally we maintain patches for our customers to download and apply to a specific version of boost, but we have also created pull requests to get some changes put back into boost trunk.

eldiener:
What I am saying is that if you provide pull requests on some Boost library, let's call it  'XXX', in order for that library to work correctly with your IBM compiler it is only natural for the maintainer of library XXX  to want to test your pull request with your IBM compiler before merging the pull request with the rest of XXX's code. Can you understand that ? If your IBM compiler is not available for the maintainer of XXX he must take it on spec that your pull request is correct. With all due respect to your own programming abilities, as well as those of your fellow IBM developers, Chris Cambly or Michael Wong, Boost developers do not in general like to rely on others saying that some change to their library is correct without being able to test that this is so. This is why I would like to suggest to you, as a representative of IBM, that it would be advantageous to your IBM compilers if Boost maintainers were able to obtain a free version of your compilers, purely for non-commerical use in testing their libraries, in order to know that any changes to their library made to accomodate your compilers are validated by testing them out.

camorton2:
Sure I understand, its a debate that has been going on for some time.

eldiener:
Then do not be surprised, or take it personally, if you submit a pull request or bug report for some Boost library pertaining to a change you want made to accommodate your IBM compiler(s), if the PR or bug report is ignored because it is difficult or onerous for the maintainer of that library to test the change. I do understand for xlcpp that you allow a 60-day trial, but that is not convenient for a Boost library developer to continually have to use, even if it were possible to do so.

Nonetheless thanks for contributing regression tests to Boost for your IBM compilers.

camorton2:
I've pointed out this discussion to Michael Wong.   I believe he was involved with similar discussions in the past and might provide some insight if any progress was made.

PR #83, opened 2015-06-11
Implement ability to have "rootless" projects.

grafikrobot:
With this it is no longer an error to not have a jamroot file when loading a project. Instead the
top-most jamfile project will operate as the jamroot.


swatanabe:
I think this is a good idea, but we need to make sure that it's actually safe.  Jamroot has some obscure differences from regular Jamfiles.  Here are the ones that I'm aware of off the top of my head:
- Jamroot is required to have project-id when --build-dir is passed on the command line
- Only Jamroot can have a build-dir when --build-dir is passed.

swatanabe:
AMDG

- In project-test3, I'd prefer to remove the initial test,
  as it no longer does anything useful.
- Can we have a more descriptive name than project-test5?
- I'm not convinced that basing the test off of project-test3/4
  is useful.  Those are primarily testing the behavior of properties,
  which is only tangentially related to what you're testing
  here.

In Christ,
Steven Watanabe

swatanabe:
AMDG

On 10/24/2018 11:59 AM, Rene Rivera wrote:
> @swatanabe I wrote totally different tests. Any thoughts?
> 

Looks good, although I personally prefer to keep
everything in the .py file instead of creating
a separate directory.

In Christ,
Steven Watanabe

PR #89, opened 2015-08-13
Free and optional features should not have default values.

frenchtoast747:


vprus:
Cherry-picked, thanks!

PR #92, opened 2015-09-03
Added boost/container

igaztanaga:
Boost.Container requires a DLL if some extended allocators are used


vprus:
Cherry-picked, thanks.

PR #95, opened 2015-09-08
Conditionally added properties should have their paths translated.

frenchtoast747:
The Jam side has the call to translate paths. Just an update to the Python Port to function the same.


vprus:
Cherry-picked, thanks!

PR #99, opened 2015-10-12
Port toolset.inherit-rules

frenchtoast747:
I'm not sure if this is the best way to handle this or not, but I believe this emulates Jam rule overriding. I had to take out the exception that is raised if an action of the same name is registered so that a user can inherit actions and then override as necessary.


vprus:
Cherry-picked, thanks!

PR #102, opened 2015-10-29
Provide error message and stacktrace with bad parameter to bjam.call()

frenchtoast747:
The previous implementation would just die when a bad parameter was passed in to `bjam.call()`.

This tries to provide an error message describing the value that was invalid and falls back to just a TypeError in the event that `repr()` cannot be called on the PyObject. Both provide a stacktrace to where the call failed.

### DIFF
--- a/src/engine/builtins.c
+++ b/src/engine/builtins.c
@@ -2089,8 +2089,19 @@ static LIST *jam_list_from_sequence(PyObject *a)
         char * s = PyString_AsString( e );
         if ( !s )
         {
-                        err_printf( "Invalid parameter type passed from Python\n" );
-            exit( 1 );
+            /* try to get the repr() on the object */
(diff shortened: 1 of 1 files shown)


frenchtoast747:
Should the Engine's version be updated when changes are made?

swatanabe:
AMDG

On 10/29/2015 08:28 AM, Aaron Boman wrote:
> Should the Engine's version be updated when changes are made?
> 

  I only change the engine version for
changes that can cause incompatibility.
e.g. adding a builtin, significant
bug fixes, etc.  This change seems
purely cosmetic.

In Christ,
Steven Watanabe

frenchtoast747:
Alright, thanks!

vprus:
Aaron,

thanks for the patch, cherry-picked.

PR #104, opened 2015-11-10
python: Determine the version if interp, incl & lib are provided

mgorny:
Fix the conditional around acquiring the data from Python interpreter to be run if interpreter-cmd, includes and libraries are provided (but version is not).


vprus:
Michał,

thanks for the patch, it has been cherry-picked.

PR #107, opened 2015-11-19
Add arm64 instruction set.

fxb:


vprus:
Hi Felix,

are you adding this new value to be used with gcc or msvc, or some other compiler? I was thinking that for gcc, it would be better to use architecture=arm + address-model=64 to select arm64 options, and possibly the same with msvc. What do you think?

fxb:
Hi Vladimir,

This is required to correctly pass "-arch arm64" to the assembler when using the "darwin" toolset. Without it, compiling Boost.Context for arm64 fails.

fxb:
Would that be under "Generic options" in darwin.jam? As long as it passes "-arch arm64" when invoking the assember, it should be fine. The reason I added the arm64 instruction set was that it was consistent with the existing way of adding the "-arch" flag in darwin.jam. Possibly setup-address-model should be refactored into the way you suggest?

vprus:
It seems to me that yes, changing setup-address-model so that the arm case checks for address-model property, and uses either 'arm' or 'arm64' will be the cleanest solution.

PR #109, opened 2015-11-19
Solaris Studio C++11 link errors

akumta:
Several python tests fail with the following error message when compiling with Oracle Solaris Studio compilers with -std=[c++03,c++11], modes.

 ImportError?: ld.so.1: isapython2.6: fatal: relocation error: file /export/home/boost_regression_develop/boost_sparc-S2_cpp11/results/boost/bin.v2/libs/python/test/injected.test/sun-next_cpp11/ release/threading-multi/injected_ext.so: symbol _ZTIv: referenced symbol not found

These errors occur when any libraries that are linked into the application that has no C++11 runtime already linked in.

When creating shared libraries, replacing '-G' with '-G -library=stdcpp,CrunG3'   seems to resolve the issue.

### DIFF
--- a/src/tools/sun.jam
+++ b/src/tools/sun.jam
@@ -148,11 +148,15 @@ actions link bind LIBRARIES
 rule link.dll ( targets * : sources * : properties * )
 {
     SPACE on $(targets) = " " ;
+    if [ MATCH "(-std=c\\+\\+[03,11])" : [ on $(targets) return $(OPTIONS) ] ]
+    {
+       STDLIBOPT on $(targets) = -library=stdcpp,CrunG3 ;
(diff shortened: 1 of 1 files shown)


vprus:
Thanks for the patch! Looking at it, I wonder whether this should be considered a compiler bug? If passing -std=c++11 results in object files that fail to work without -library=stdcpp,GrunG3, then I would imagine the compiler or linker should automatically link to that newer library? Is this something you can check with compiler vendor, before we try to work around this issue?

akumta:
Yes this is a known compiler issue and the patch here is a workaround for it until the compiler issue is 
fixed.

PR #114, opened 2015-12-14
Run tests via Travis CI on OS X.

tee3:
This pull request runs the tests on OS X via Travis CI.  The tests are now passing.

See https://github.com/boostorg/build/pull/110 for initial discussion of this branch.  The issues Travis CI had previously are no longer an issue for building the Boost.Build engine under Travis CI's OS X support.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,16 @@
 sudo: false
 os:
   - linux
+  - osx
 env:
   - TOOLSET=gcc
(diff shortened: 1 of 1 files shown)


tee3:
Even though this is passing, I'd like to defer it until after the actual tests are running against Linux.  The OS X support for Python is not available yet and I don't know when it will be.

Sorry about the noise!

PR #117, opened 2015-12-23
Auto configuring MPI on Windows

praetorian20:
Microsoft has a new MPI implementation available - [Microsoft MPI](https://msdn.microsoft.com/en-us/library/windows/desktop/bb524831%28v=vs.85%29.aspx). There are now two installers, one for the MPI executor and another for the MPI SDK (which contains the headers and static libraries), and these install to different locations by default.

The jam file has been changed so that MS-MPI is first checked, if found use that as the MPI executor/SDK. If not found, the MS Compute Cluster Pack is checked next.

The changes were tested on Windows 10 x64 with VS2015.

Also stripped trailing whitespace from the file.

### DIFF
--- a/src/tools/mpi.jam
+++ b/src/tools/mpi.jam
@@ -245,15 +245,47 @@ rule init ( mpicxx ? : options * : mpirun-with-options * )
 
   if ! $(mpicxx) && [ os.on-windows ]
   {
-    # Try to auto-configure to the Microsoft Compute Cluster Pack
+    # Paths for Microsoft MPI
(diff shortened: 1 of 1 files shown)


jhunold:
Tested the patch with the latest Microsoft MPI and Visual Studio 2015. Works. 
I'd like to merge this, but "Also stripped trailing whitespace from the file." is obfuscating the changes. Can you please split this into two commits? One removing the whitespace first and then making the real changes. Otherwise it is really hard to figure out what has changed.

praetorian20:
Done. Also made the MPI changes commit message a lot more descriptive.

jhunold:
Thanks for the cleanup.

PR #121, opened 2016-02-22
Fix ImportError message.

frenchtoast747:
It's possible that when importing a module (that exists and is importable) the module itself contains an ImportError. This is caught by the exception handling and falsely reports that the module itself is not able to be found.

This change makes sure that ImportError is referring to the module trying to be imported. If it's not, then the original ImportError is re-raised in order to inform the end-user correctly.

### DIFF
--- a/src/build/project.py
+++ b/src/build/project.py
@@ -730,10 +730,21 @@ actual value %s""" % (jamfile_module, saved_project, self.current_project))
             module = imp.load_module(mname, f, location, description)
             self.loaded_tool_modules_[name] = module
             return module
-        except ImportError:
+        except ImportError as e:
+            # it's possible that the module being imported exists and
+            # is able to be imported, but contains an ImportError. This
+            # check makes sure that the error message pertains to the
+            # module that *we* are trying to import so that the proper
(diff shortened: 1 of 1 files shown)


swatanabe:
AMDG

  This seems rather fragile.  Is there any
way to determine this without parsing the
string?  Does python explicitly guarantee
the format of the string?  This would fail
if the error message could be localized,
for instance.

In Christ,
Steven Watanabe

frenchtoast747:
You're right, this is terrible.

Ideally, the module that is not importable should be set as an attribute on the `ImportError` instance, but it's not. There was [a patch for this functionality](https://bugs.python.org/issue1559549), but it's only found in Python 3.3+

It might be possible to get the module that failed from the stack, but I think there might be some parsing involved with that as well.

I'll continue research on this.

vprus:
Aaron, is this still outstanding? <br><br>:octocat: *Sent from [GH](http://nodegh.io).*

frenchtoast747:
Yes, it still is. I did not include this change on the features/python-updates branch as it's still a nasty hack.

PR #123, opened 2016-02-29
Add support for FreeBSD 10 or later.

Flast:
Since FreeBSD 10, it uses clang as a system default compiler.

### DIFF
--- a/src/engine/build.sh
+++ b/src/engine/build.sh
@@ -77,6 +77,7 @@ Guess_Toolset ()
           BOOST_JAM_TOOLSET=vacpp
        fi
     elif test_uname AIX && test_path xlc; then BOOST_JAM_TOOLSET=vacpp    
+    elif test_uname FreeBSD && test_path freebsd-version; then BOOST_JAM_TOOLSET=clang
     elif test_path gcc ; then BOOST_JAM_TOOLSET=gcc
     elif test_path icc ; then BOOST_JAM_TOOLSET=intel-linux
     elif test -r /opt/intel/cc/9.0/bin/iccvars.sh ; then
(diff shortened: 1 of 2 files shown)


src/engine/build.sh
```diff
@@ -77,6 +77,7 @@ Guess_Toolset ()
           BOOST_JAM_TOOLSET=vacpp
        fi
     elif test_uname AIX && test_path xlc; then BOOST_JAM_TOOLSET=vacpp    
+    elif test_uname FreeBSD && test_path freebsd-version; then BOOST_JAM_TOOLSET=clang
```
grafikrobot: Is this is for FreeBSD >=10.. Shouldn't the version also be checked?

swatanabe: AMDG

On 02/29/2016 08:37 AM, Rene Rivera wrote:
>> @@ -77,6 +77,7 @@ Guess_Toolset ()
>>            BOOST_JAM_TOOLSET=vacpp
>>         fi
>>      elif test_uname AIX && test_path xlc; then BOOST_JAM_TOOLSET=vacpp    
>> +    elif test_uname FreeBSD && test_path freebsd-version; then BOOST_JAM_TOOLSET=clang
> 
> Is this is for FreeBSD >=10.. Shouldn't the version also be checked?
> 

  Also, shouldn't this be testing whether clang exists?

In Christ,
Steven Watanabe

Flast: >> Is this is for FreeBSD >=10.. Shouldn't the version also be checked?

`freebsd-version` is new command which is provided only if FreeBSD >= 10.

> Also, shouldn't this be testing whether clang exists?

OK, I`ll check `test_path clang`.

vprus:
Is the current version of the patch fine with everybody? I'd rather have it in 1.61 if so.

Flast:
I have no more patch.

vprus:
I've cherry-picked this pull request, thanks!

PR #127, opened 2016-04-29
Documentation fixes, largely grammar and spelling corrections.

eldiener:


doc/src/overview.xml
```diff
@@ -675,10 +675,10 @@ b2 toolset=gcc variant=debug optimization=space
             <listitem>
               <para>Changes the build directories for all project roots being built. When
               this option is specified, all Jamroot files must declare a project name.
-              The build directory for the project root will be computed by concatanating
+              The build directory for the project root will be computed by concatenating
               the value of the <option>--build-dir</option> option, the project name
               specified in Jamroot, and the build dir specified in Jamroot
-              (or <literal>bin</literal>, if none is specified).
+              (or <literal>bin.v2</literal>, if none is specified).
```
swatanabe: The original version is correct.  bin.v2 is specified by the superproject's Jamroot.

doc/src/reference.xml
```diff
@@ -561,13 +561,15 @@ path-constant DATA : data/a.txt ;
       <varlistentry>
         <term><literal>cflags</literal></term>
         <term><literal>cxxflags</literal></term>
+        <term><literal>compileflags</literal></term>
```
swatanabe: compileflags is not a feature.  What you're thinking of is the use of compileflags here: http://www.boost.org/build/doc/html/bbv2/reference/tools.html, which is different.  I realize that this is confusing, but the existing documentation does describe the current behavior correctly.

doc/src/reference.xml
```diff
@@ -686,8 +688,8 @@ path-constant DATA : data/a.txt ;
         <listitem>
           <para><emphasis role="bold">Allowed values:</emphasis> <literal>on</literal>, <literal>off</literal>.</para>
 
-          <para>The <literal>runtime-debugging</literal> feature specifies if
-          produced object files, executables and libraries should include
+          <para>The <literal>runtime-debugging</literal> feature specifies that,
+          if produced, object files, executables, and libraries should include
```
swatanabe: s/if/whether/ would be a more accurate rendering of the original intent of this sentence.

doc/src/tasks.xml
```diff
@@ -530,7 +530,7 @@ unit-test helpers_test
 </programlisting>
             The command used to run the executable will be:
 <screen>
-<emphasis role="bold">valgrind</emphasis> bin/$toolset/debug/helpers_test
+<emphasis role="bold">valgrind</emphasis> bin.v2/$toolset/debug/helpers_test
```
swatanabe: Again, bin.v2 is specific to Boost.  The generic default is bin.

doc/src/tutorial.xml
```diff
@@ -280,8 +280,8 @@ top/
       in their requirements, too.  Of course, any project can add
       include paths to those specified by its parents. <footnote>
       <para>Many
-      features will be overridden,
-      rather than added-to, in subprojects.  See <xref
+      features will be overridden in,
+      rather than added-to, subprojects.  See <xref
```
swatanabe: The original formulation makes more sense.

doc/src/reference.xml
eldiener: I do not see how writing that 'cflags' applies to both C and C++ can be correct. Clearly this is not the case for any compiler toolset that is documented. Leaving that in the documentation is just a cause for confusion. It had certainly confused me until I realized it wasn't tthe case for any compiler toolset which is documented in Boost Build.

doc/src/tutorial.xml
eldiener: I disagree. The phrase 'added-to in subprojects' cannot be correct English while the phrase 'added-to subprojects' is the correct terminology.

doc/src/overview.xml
eldiener: I will make the correction.

doc/src/reference.xml
eldiener: I see now that you are correct and that 'specified whether produced object files...' etc. is the correct terminology. I will make the change.

doc/src/tasks.xml
eldiener: I will make the change.

doc/src/tutorial.xml
swatanabe: AMDG

On 04/29/2016 10:44 AM, Edward Diener wrote:
>> @@ -280,8 +280,8 @@ top/
>>        in their requirements, too.  Of course, any project can add
>>        include paths to those specified by its parents. <footnote>
>>        <para>Many
>> -      features will be overridden,
>> -      rather than added-to, in subprojects.  See <xref
>> +      features will be overridden in,
>> +      rather than added-to, subprojects.  See <xref
> 
> I disagree. The phrase 'added-to in subprojects' cannot be correct English while the phrase 'added-to subprojects' is the correct terminology.
> 

It's the features that are being added-to, not the subproject.

In Christ,
Steven Watanabe

doc/src/reference.xml
swatanabe: AMDG

On 04/29/2016 10:41 AM, Edward Diener wrote:
>> @@ -561,13 +561,15 @@ path-constant DATA : data/a.txt ;
>>        <varlistentry>
>>          <term><literal>cflags</literal></term>
>>          <term><literal>cxxflags</literal></term>
>> +        <term><literal>compileflags</literal></term>
> 
> I do not see how writing that 'cflags' applies to both C and C++ can be correct.

Sadly it is.

In Christ,
Steven Watanabe

doc/src/tutorial.xml
eldiener: It's perfectly valid in the main clause to say:
`Many features will be overridden in subprojects.`
but saying in the alternative clause:
`Many features will be added-to in subprojects`
is not as correct as saying:
`Many features will be added-to subprojects.`

If you do not see this I will change it back and not try to correct it.

doc/src/reference.xml
eldiener: I will remove the change but I strongly feel that the documentation, even if correct in general, will just serve to lead end-users of Boost Build astray. At the very least a further explanation should be added directing the end-user to look at the toolset for his compiler where the meaning of 'cflag' often will be different.

eldiener: I do not understand how 'cflags' can mean command line options for both C and C++ when the various compiler toolsets clearly state that 'cflags' specifies command line options for C mode and that 'compileflags' specifies command line options for both C and C++ mode. Can you explain this to me ?

swatanabe: AMDG

On 04/29/2016 11:14 AM, Edward Diener wrote:
>> @@ -561,13 +561,15 @@ path-constant DATA : data/a.txt ;
>>        <varlistentry>
>>          <term><literal>cflags</literal></term>
>>          <term><literal>cxxflags</literal></term>
>> +        <term><literal>compileflags</literal></term>
> 
> I will remove the change but I strongly feel that the documentation, even if correct in general, will just serve to lead end-users of Boost Build astray. At the very least a further explanation should be added directing the end-user to look at the toolset for his compiler where the meaning of 'cflag' often will be different.
> 

  No that's wrong.  The meaning of cflags as a
feature is consistent.  cflags does apply to
both C and C++ for all toolsets.  I checked
both msvc and gcc:

toolset.flags gcc.compile USER_OPTIONS <cflags> ;
toolset.flags msvc.compile OPTIONS <cflags> : ;

In Christ,
Steven Watanabe

swatanabe: AMDG

On 04/29/2016 01:18 PM, Edward Diener wrote:
>> @@ -561,13 +561,15 @@ path-constant DATA : data/a.txt ;
>>        <varlistentry>
>>          <term><literal>cflags</literal></term>
>>          <term><literal>cxxflags</literal></term>
>> +        <term><literal>compileflags</literal></term>
> 
> I do not understand how 'cflags' can mean command line options for both C and C++ when the various compiler toolsets clearly state that 'cflags' specifies command line options for C mode and that 'compileflags' specifies command line options for both C and C++ mode. Can you explain this to me ?
> 

  You're confusing cflags as an option for
toolset initialization with cflags as a feature.
The arguments for 'using xxx' are toolset specific
and are essentially named parameters.  They are
completely independent from the feature/property
framework, even though they share the same syntax.

In Christ,
Steven Watanabe

eldiener: OK, I understand the distinction now, and where my previous confusion occurred.

Having the exact same feature name meaning different things depending on where it is used seems to me a really bad programming design decision for Boost Build.

I will remove the change I made, realizing it is wrong.

eldiener:
Is there any objections to pushing these documentation fixes now ?

vprus:
Steven, do you have any further comments?

swatanabe:
AMDG

On 05/17/2016 03:28 AM, Vladimir Prus wrote:
> Steven, do you have any further comments?
> 

No, I think it's good.

In Christ,
Steven Watanabe

vprus:
I've merged this pull request, thanks!

PR #131, opened 2016-06-21
Add Apple TV support

vmrob:
The bulk of the patch is a copy/paste job from iphoneos with the additional SDKs for Apple TV.


vprus:
Cherry-picked this one, thanks!

PR #132, opened 2016-07-03
Add support for Visual Studio 15

MarcelRaad:
Tested with Visual Studio 15 Preview 2. The compiler itself is the same as with Visual Studio 2015, but if only Visual Studio 15 is installed, the compiler could not be found.

### DIFF
--- a/src/engine/build.bat
+++ b/src/engine/build.bat
@@ -28,7 +28,7 @@ ECHO ### You can specify the toolset as the argument, i.e.:
 ECHO ###     .\build.bat msvc
 ECHO ###
 ECHO ### Toolsets supported by this script are: borland, como, gcc, gcc-nocygwin,
-ECHO ###     intel-win32, metrowerks, mingw, msvc, vc7, vc8, vc9, vc10, vc11, vc12, vc14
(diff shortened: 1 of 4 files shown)


jhunold:
Something is wrong with this patch. The compiler version for "Visual Studio 2015" is "vc14" and supported for a while. Building with only "Visual Studio 2015" installed works for me.

MarcelRaad:
@jhunold This is not for VS 14 (2015), but for VS 15 (2017?). It currently comes with the same compiler as VS 14 in the Preview, but if you only have VS 15 installed, building Boost is currently not possible.

If you're confused by the comment in build.jam, it's above the vc14 section.

jhunold:
Then please update the commit message accordingly. Something along the lines of "Add support for upcoming Visual Studio 2016(/17)" and "Tested with Visual Studio 16(/17) Preview 2." 
We have to distinguish between the IDE (Visual Studio 2016(/17)) and the compiler (vc15) here. 
It would be best if you can figure out the final IDE version as I'm not following the msvc development that closely.
Otherwise the patch looks good to me.

MarcelRaad:
Thanks! I've changed the commit message. Microsoft themselves always call it Visual Studio "15" Preview. I'm still hoping they'll finally get rid of the double numbering with version number and year and just call it VS 15.

PR #134, opened 2016-08-05
Create README and CONTRIBUTING.

tee3:
This pull request creates reStructuredText documents for README and CONTRIBUTING.  This makes the project a little more friendly for developers using a hosting service (such as GitHub).  The README is displayed nicely by default and the CONTRIBUTING document is highlighted when creating an issue or a pull request.

The README just currently points to the website and the CONTRIBUTING document, but could grow to be more informative if desired.

The CONTRIBUTING document is the original hacking document, minimally converted to reStructuredText to display nicely by hosting services that support this functionality.  This could and probably should be updated to help developers make changes, but this makes no changes at all to the content of the document.

This pull request goes through the process of adding these files in detail, so it could be squashed to make it smaller.  I happen to think the small diffs at each step help in review.


tee3:
Thanks @vprus!

PR #136, opened 2016-08-15
Fix strip-eol for long command output

vmrob:
Previously, when shell command output exceeded the 1024 character
buffer size, each chunk of output would be stripped. This had the
undesirable effect of sometimes breaking compilation by splitting on
whitespace boundaries. This patch addresses the issue by providing a
string_rtrim function and utilizing the function on the output string
instead of each buffered component.

### DIFF
--- a/src/engine/builtins.c
+++ b/src/engine/builtins.c
@@ -450,7 +450,7 @@ void load_builtins()
         char const * args [] = { "path", 0 };
         bind_builtin( "MAKEDIR", builtin_makedir, 0, args );
     }
-    
+
     {
         const char * args [] = { "path", 0 };
         bind_builtin( "READLINK", builtin_readlink, 0, args );
@@ -1981,7 +1981,7 @@ LIST *builtin_readlink( FRAME * frame, int flags )
         bufsize *= 2;
         buf = BJAM_MALLOC( bufsize );
     }
-    
+
     if ( buf != static_buf )
         BJAM_FREE( buf );
 
@@ -2460,15 +2460,6 @@ PyObject * bjam_caller( PyObject * self, PyObject * args )
 #endif  /* defined(_MSC_VER) || defined(__BORLANDC__) */
 
 
-static char * rtrim( char * const s )
-{
-    char * p = s;
-    while ( *p ) ++p;
-    for ( --p; p >= s && isspace( *p ); *p-- = 0 );
-    return s;
-}
-
-
 LIST * builtin_shell( FRAME * frame, int flags )
 {
     LIST   * command = lol_get( frame->args, 0 );
@@ -2515,14 +2506,15 @@ LIST * builtin_shell( FRAME * frame, int flags )
         buffer[ ret ] = 0;
         if ( !no_output_opt )
         {
-            if ( strip_eol_opt )
-                rtrim( buffer );
             string_append( &s, buffer );
         }
 
         /* Explicit EOF check for systems with broken fread */
         if ( feof( p ) ) break;
     }
+
+    if ( strip_eol_opt )
+        string_rtrim( &s );
 
     exit_status = pclose( p );
 
--- a/src/engine/strings.c
+++ b/src/engine/strings.c
@@ -189,6 +189,12 @@ char string_back( string * self )
     return self->value[ self->size - 1 ];
 }
 
+void string_rtrim( string * self )
+{
+    assert_invariants( self );
+    char * p = self->value + self->size - 1;
+    for ( p; p >= self->value && ( *p == '\0' || isspace( *p ) ); *p-- = 0 );
+}
 
 #ifndef NDEBUG
 void string_unit_test()
@@ -218,6 +224,26 @@ void string_unit_test()
         assert( !strcmp( copy->value, original ) );
         assert( copy->size == strlen( original ) );
         string_free( copy );
+    }
+
+    {
+        char * const foo = "Foo    ";
+        string foo_copy[ 1 ];
+        string_copy( foo_copy, foo );
+        string_rtrim( foo_copy );
+        assert( !strcmp( foo_copy->value, "Foo" ) );
+
+        string_rtrim( foo_copy );
+        assert( !strcmp( foo_copy->value, "Foo" ) );
+
+        char * const bar = "Bar\0\0\0";
+        string bar_copy[ 1 ];
+        string_copy( bar_copy, bar );
+        string_rtrim( bar_copy );
+        assert( !strcmp( bar_copy->value, "Bar" ) );
+
+        string_rtrim( bar_copy );
+        assert( !strcmp( bar_copy->value, "Bar" ) );
     }
 }
 #endif
--- a/test/core-language/test.jam
+++ b/test/core-language/test.jam
@@ -1239,7 +1239,7 @@ for y in $(values)
 }
 
 check-order break-for-exec : r1-v1 ;
-check-equal break-for-cleanup : $(z) : original ; 
+check-equal break-for-cleanup : $(z) : original ;
 
 for local y in $(values)
 {
@@ -1250,7 +1250,7 @@ for local y in $(values)
 }
 
 check-order break-for-local-exec : r1-v1 ;
-check-equal break-for-local-cleanup : $(z) : original ; 
+check-equal break-for-local-cleanup : $(z) : original ;
 
 local z1 = z1val ;
 local z2 = z2val ;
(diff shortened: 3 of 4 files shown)


vmrob:
Please don't merge this until we can confirm that the `string_unit_test()` function runs properly. I was unable to determine how run the C-based unit tests and only confirmed that `python core_language.py` worked on my machine.

vmrob:
Hi @vprus, any thoughts on merging this patch?

vprus:
Hi Victor,

which system was you testing on? Both on Unbuntu and Windows, the unit test appears to pass even without the code patch ;-)

vmrob:
Very interesting! Please note that I had to add a new test to demonstrate the bug.

I was previously testing on MacOS 10.11 and now MacOS 10.12. I can't easily downgrade my OS, but the following is what I see with test output before/after the patch **using the new test**:

```
$ python test_all.py darwin
Note: skipping extra tests
unit_tests                        : PASSED
module_actions                    : PASSED
startup_v2                        : PASSED
core_d12                          : PASSED
core_typecheck                    : PASSED
core_delete_module                : PASSED
core_language                     : FAILED
core_arguments                    : PASSED
core_varnames                     : PASSED
core_import_module                : PASSED
absolute_sources                  : PASSED
alias                             : PASSED
alternatives                      : PASSED
bad_dirname                       : PASSED
build_dir                         : PASSED
build_file                        : PASSED
build_no                          : PASSED
builtin_echo                      : PASSED
builtin_exit                      : PASSED
builtin_glob                      : PASSED
builtin_glob_archive              : Set environmental variable 'DO_DIFF' to examine the difference.
FAILED
builtin_split_by_characters       : PASSED
bzip2                             : PASSED
c_file                            : PASSED
chain                             : PASSED
clean                             : PASSED
composite                         : PASSED
conditionals                      : PASSED
conditionals2                     : PASSED
conditionals3                     : PASSED
conditionals_multiple             : PASSED
configuration                     : PASSED
copy_time                         : PASSED
core_action_output                : PASSED
core_action_status                : PASSED
core_actions_quietly              : PASSED
core_at_file                      : PASSED
core_bindrule                     : PASSED
core_jamshell                     : PASSED
core_multifile_actions            : PASSED
core_nt_cmd_line                  : PASSED
core_option_d2                    : PASSED
core_option_l                     : PASSED
core_option_n                     : PASSED
core_parallel_actions             : PASSED
core_parallel_multifile_actions_1 : PASSED
core_parallel_multifile_actions_2 : PASSED
core_source_line_tracking         : PASSED
core_update_now                   : PASSED
core_variables_in_actions         : PASSED
custom_generator                  : PASSED
default_build                     : PASSED
default_features                  : PASSED
dependency_property               : PASSED
dependency_test                   : PASSED
direct_request_test               : PASSED
disambiguation                    : PASSED
dll_path                          : PASSED
double_loading                    : PASSED
duplicate                         : PASSED
example_libraries                 : PASSED
example_make                      : PASSED
exit_status                       : PASSED
expansion                         : PASSED
explicit                          : PASSED
feature_cxxflags                  : PASSED
free_features_request             : PASSED
generator_selection               : PASSED
generators_test                   : PASSED
implicit_dependency               : PASSED
indirect_conditional              : PASSED
inherit_toolset                   : PASSED
inherited_dependency              : PASSED
inline                            : PASSED
lib_source_property               : PASSED
library_chain                     : PASSED
library_property                  : PASSED
link                              : PASSED
load_order                        : PASSED
loop                              : PASSED
make_rule                         : PASSED
message                           : PASSED
ndebug                            : PASSED
no_type                           : PASSED
notfile                           : PASSED
ordered_include                   : PASSED
out_of_tree                       : PASSED
path_features                     : PASSED
prebuilt                          : PASSED
print                             : PASSED
project_dependencies              : PASSED
project_glob                      : PASSED
project_id                        : PASSED
project_root_constants            : PASSED
project_root_rule                 : PASSED
project_test3                     : PASSED
project_test4                     : PASSED
property_expansion                : PASSED
rebuilds                          : PASSED
regression                        : PASSED
relative_sources                  : PASSED
remove_requirement                : PASSED
rescan_header                     : PASSED
resolution                        : PASSED
scanner_causing_rebuilds          : PASSED
searched_lib                      : PASSED
skipping                          : PASSED
sort_rule                         : PASSED
source_locations                  : PASSED
source_order                      : PASSED
space_in_path                     : PASSED
stage                             : PASSED
standalone                        : PASSED
static_and_shared_library         : PASSED
suffix                            : PASSED
tag                               : PASSED
test_result_dumping               : PASSED
test_rc                           : PASSED
testing_support                   : PASSED
timedata                          : PASSED
toolset_requirements              : PASSED
unit_test                         : PASSED
unused                            : PASSED
use_requirements                  : PASSED
using                             : PASSED
wrapper                           : PASSED
wrong_project                     : PASSED
zlib                              : PASSED
symlink                           : PASSED

        === Test summary ===
        PASS: 127
        FAIL: 2
```

I'll push an intermediate revision without the fix to demonstrate the broken behavior on Travis.

vmrob:
The above [travis builds](https://travis-ci.org/boostorg/build/builds/166141860) appears to be failing as a result of the bug. Note the failure in `core_language`.

vmrob:
I've rebased this patch on top of the latest develop. @vprus have you had any time to review the patch? See the inline comments on the commit for a more detailed breakdown.

src/engine/builtins.c
```diff
@@ -2515,15 +2506,16 @@ LIST * builtin_shell( FRAME * frame, int flags )
         buffer[ ret ] = 0;
         if ( !no_output_opt )
         {
-            if ( strip_eol_opt )
-                rtrim( buffer );
```
vmrob: Previously, this would occur _inside_ the loop which is split on 1024 byte boundaries in commands.

src/engine/builtins.c
```diff
@@ -2515,15 +2506,16 @@ LIST * builtin_shell( FRAME * frame, int flags )
...
             string_append( &s, buffer );
         }
 
         /* Explicit EOF check for systems with broken fread */
         if ( feof( p ) ) break;
     }
 
+    if ( strip_eol_opt )
+        string_rtrim( &s );
```
vmrob: By moving it outside the loop and only processing the final string, the bug is mitigated.

src/engine/strings.c
```diff
@@ -219,5 +225,25 @@ void string_unit_test()
...
+
+        char * const bar = "Bar\0\0\0";
+        string bar_copy[ 1 ];
+        string_copy( bar_copy, bar );
+        string_rtrim( bar_copy );
+        assert( !strcmp( bar_copy->value, "Bar" ) );
+
+        string_rtrim( bar_copy );
+        assert( !strcmp( bar_copy->value, "Bar" ) );
+    }
```
vmrob: I do not know how to run these tests. I'm reasonably confident they work, but I do not know how to verify this. **How can I verify that these tests function as expected?**

test/core-language/test.jam
```diff
@@ -1505,11 +1505,32 @@ check-equal shell : "value\n"   : [ SHELL $(c) ] ;
...
 
+# buffered output
+
+local expected = "When the shell output buffer splits on whitespace, the whitespace shouldn't be trimmed.                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                             end." ;
```
vmrob: This string is the key component to these tests. It's hard to see, but there is a large amount of whitespace between `trimmed.` and `end.`. The length of this string is 1096 bytes. Provided that the output is split in 1024 byte regions, the first chunk will have a large amount of whitespace trailing after `trimmed.`.

The more common error case (as I've seen), is when trying to parse something like the output of a `pkg-config` command and the whitespace boundary just _happens_ to be between a set of cflags or linker flags.

vmrob:
@vprus Have you had a chance to review this patch?

vmrob:
@eldiener I noticed you are an active contributor to this project. Do you think you could take a look at this PR? Thanks!

eldiener:
Testing out your PR I am seeing the core_language test failing, whereas it normally succeeds on the 'develop' branch. I using msvc-14.0 on Windows.

vmrob:
@eldiener Just tested this patch rebased against develop (1949282) and core_language seems to work on Darwin... I don't have quick access to MSVC to test this with, but I'm betting there's something in the shell command that isn't transferring correctly.

It relies on parsing the output of an echo command exceeding 1024 characters:

```
echo "When the shell output buffer splits on whitespace, the whitespace shouldn't be trimmed.                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                             end."
```

Any ideas?

eldiener:
If you run  Windows you can get a free version of VS2015 with which to test. If you do not run Windows then of course the free version does you no good. I will try to test on Linux to see what my results are. I do not have a Mac.

eldiener:
Sorry for the delay with this. I did test under Linux and the tests passed with your fix. I now have to try to figure out why the core_language test is failing with VC++ under Windows.

swatanabe:
AMDG

On 05/02/2017 06:08 PM, Edward Diener wrote:
> Sorry for the delay with this. I did test under Linux and the tests passed with your fix. I now have to try to figure out why the core_language test is failing with VC++ under Windows.
> 

  It fails because echo behaves differently on Windows
(the quotes appear in the output).

In Christ,
Steven Watanabe

test/core-language/test.jam
```diff
@@ -1505,11 +1505,32 @@ check-equal shell : "value\n"   : [ SHELL $(c) ] ;
...
+local buffered = "echo \"$(expected)\"" ;
+if $(OS) = VMS { buffered = "PIPE WRITE SYS$OUTPUT \"$(expected)\"" ; }
```
eldiener: You need to add this line after 1518:

`if $(OS) = NT { buffered = "echo $(expected)" ; }`

That addition fixes the problem where your test does not pass under Windows.

eldiener:
Thanks Steven. I suggested a fix for the PR that works in my testing.

eldiener:
I have tested this fix on both Windows and Linux and it works. After merging I will update this fix with the fix I specified above for Windows and push it to develop.

eldiener:
I am not able to merge this PR online. I do not know why but I suspect it may have conflicts with the base branch.

eldiener:
I pushed this fix to develop on my own, along with my own fix for Windows.

PR #158, opened 2017-02-01
Some fixes for VS2017

KindDragon:
Path autodetection to VS works only if it installed to default path

### DIFF
--- a/src/engine/build.bat
+++ b/src/engine/build.bat
@@ -106,9 +106,19 @@ if NOT "_%VS150COMNTOOLS%_" == "__" (
     set "BOOST_JAM_TOOLSET_ROOT=%VS150COMNTOOLS%..\..\VC\"
     goto :eof)
 call :Clear_Error
-if EXIST "%ProgramFiles%\Microsoft Visual Studio 15.0\VC\VCVARSALL.BAT" (
+if EXIST "%ProgramFiles%\Microsoft Visual Studio\2017\Enterprise\VC\Auxiliary\Build\vcvarsall.bat"  (
     set "BOOST_JAM_TOOLSET=vc15"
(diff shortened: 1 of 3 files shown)


KindDragon:
Is it acceptable to call this PowerShell snippet http://stackoverflow.com/a/41968736/61505 from Python?
Also, we can create a small C++ tool that will return path to VS2017 and distribute it binary with Boost.Build.
IDK how to do that in plain Python

src/engine/build.bat
```diff
@@ -460,7 +470,7 @@ if NOT "_%BOOST_JAM_TOOLSET%_" == "_vc15_" goto Skip_VC15
 if NOT "_%VS150COMNTOOLS%_" == "__" (
     set "BOOST_JAM_TOOLSET_ROOT=%VS150COMNTOOLS%..\..\VC\"
     )
-if "_%VCINSTALLDIR%_" == "__" call :Call_If_Exists "%BOOST_JAM_TOOLSET_ROOT%VCVARSALL.BAT" %BOOST_JAM_ARGS%
+if "_%VCINSTALLDIR%_" == "__" call :Call_If_Exists "%BOOST_JAM_TOOLSET_ROOT%Auxiliary\Build\vcvarsall.bat" %BOOST_JAM_ARGS%
```
grafikrobot: Won't that break with earlier VS versions as they don't have the "Auxiliary\Build" path?

KindDragon: I thought that the line https://github.com/KindDragon/build/blob/cb70dbeb5cadfe94b48ad0d97452811cc4e5bbfc/src/engine/build.bat#L458  for this

grafikrobot: Ah, right.. Didn't realize the context. NM then.

KindDragon:
@grafikrobot @frenchtoast747 @vprus What do you think about my question? Now it is impossible to get the path to VS2017 without access to the VS installer COM object.

Issue #157, opened 2017-01-31
Support VS2017


KindDragon:
This patch https://github.com/boostorg/build/commit/ae5e63a1315493dfb2fc60c4d537674fa637f898 doesn't work with VS2017RC1 

See also [Finding the location of a VC++2017 install](https://blogs.msdn.microsoft.com/vcblog/2016/10/07/compiler-tools-layout-in-visual-studio-15/)

BTW toolset should be vc141 for VS2017 and %VSxxxCOMNTOOLS% environment variable doesn't exist

KindDragon:
How execute command `powershell -v 3 -exec bypass .\return-msvc-path.ps1 15` in jam file and get their output and exitcode?

KindDragon:
Script: https://github.com/KindDragon/build/commit/c03c11cc338ba2dad9be6f4aaec74fd3212dec84#diff-edcbda45192bb7d102a378db2cabd8f3

MarcelRaad:
> %VSxxxCOMNTOOLS% environment variable doesn't exist

`%VS150COMNTOOLS%` does exist if you use one of the Visual Studio command prompts.

KindDragon:
Yes, but for previous versions VS this was not required. Not convenient to run the VS command prompts before Boost.Build

AndrewPardoe:
There are three solutions to this problem:

1. Store a small executable in your repo. Run that executable to locate VS installs. You can identify the latest VS install, and ascertain that it contains MSVC (as opposed to a C#-only install.) We can help you create the optimal (as in, smallest size and specific operation) executable. 

2. Use a PowerShell module. This offers the same capability as option 1, but requires that you install a PS module on the user's machine. Worse, I think that the install is interactive (though non-admin.) On Win10 it's straightforward, on earlier versions it's more difficult to install PS modules. More details here: https://blogs.msdn.microsoft.com/heaths/2017/01/25/visual-studio-setup-powershell-module-available/, invoking the command as so: Get-VSSetupInstance | Select-VSSetupInstance -Latest -Require Microsoft.VisualStudio.Component.VC.Tools.x86.x64 

3. Brute force it through a command script. The following command script searches every single-lettered drive (whether or not it exists) for a file called vcvars64.bat. It calls every vcvars64 it finds. It then checks to make sure it can find a cl.exe. It's unholy and I'm ashamed to have written it but it may suit your purposes:

[...]

PR #159, opened 2017-02-02
Fixed paths for vc15 and arch variable

teeks99:
I had some problems when my tester with Visual Studio 2017 RC (msvc-15.0) installed pulled the boost build commit that included merge #158. 

Specifically, the default install path for msvc-15.0 is `C:\Program Files (x86)\...` so if we are going to use the paths to determine if it is installed (that is super dangerous!) we should use the `%ProgramFiles(x86)%` env variable. When building on a 32-bit system (or from a 32-bit command prompt like you get when python 32-bit calls `os.system('build.bat')`) that variable with simply point to the same location as `%ProgramFiles%`, so I'd argue it is better to use this.

Then there was an issue with the call to `vsvarsall.bat`, which in msvc-15.0 has a required first argument of the architecture. I'm not sure if there is a canonical boost  build way to pick a default architecture...so I start with x86, then use the `%PROCESSOR_ARCHTECTURE%` variable to upgrade to amd64 for most users. However if the user's environment includes the  `%Platform%` variable, it will override the others...as this is what microsoft sets up when they are doing cross-compiling (e.g. amd64 -> x86).

### DIFF
--- a/src/engine/build.bat
+++ b/src/engine/build.bat
@@ -106,19 +106,19 @@ if NOT "_%VS150COMNTOOLS%_" == "__" (
     set "BOOST_JAM_TOOLSET_ROOT=%VS150COMNTOOLS%..\..\VC\"
     goto :eof)
 call :Clear_Error
-if EXIST "%ProgramFiles%\Microsoft Visual Studio\2017\Enterprise\VC\Auxiliary\Build\vcvarsall.bat"  (
+if EXIST "%ProgramFiles(x86)%\Microsoft Visual Studio\2017\Enterprise\VC\Auxiliary\Build\vcvarsall.bat"  (
     set "BOOST_JAM_TOOLSET=vc15"
-    set "BOOST_JAM_TOOLSET_ROOT=%ProgramFiles%\Microsoft Visual Studio\2017\Enterprise\VC\"
+    set "BOOST_JAM_TOOLSET_ROOT=%ProgramFilesi(x86)%\Microsoft Visual Studio\2017\Enterprise\VC\"
     goto :eof)
 call :Clear_Error
-if EXIST "%ProgramFiles%\Microsoft Visual Studio\2017\Professional\VC\Auxiliary\Build\vcvarsall.bat"  (
+if EXIST "%ProgramFilesi(x86)%\Microsoft Visual Studio\2017\Professional\VC\Auxiliary\Build\vcvarsall.bat"  (
     set "BOOST_JAM_TOOLSET=vc15"
-    set "BOOST_JAM_TOOLSET_ROOT=%ProgramFiles%\Microsoft Visual Studio\2017\Professional\VC\"
+    set "BOOST_JAM_TOOLSET_ROOT=%ProgramFilesi(x86)%\Microsoft Visual Studio\2017\Professional\VC\"
     goto :eof)
 call :Clear_Error
-if EXIST "%ProgramFiles%\Microsoft Visual Studio\2017\Community\VC\Auxiliary\Build\vcvarsall.bat"  (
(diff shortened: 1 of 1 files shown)


src/engine/build.bat
```diff
@@ -106,19 +106,19 @@ if NOT "_%VS150COMNTOOLS%_" == "__" (
     set "BOOST_JAM_TOOLSET_ROOT=%VS150COMNTOOLS%..\..\VC\"
     goto :eof)
 call :Clear_Error
-if EXIST "%ProgramFiles%\Microsoft Visual Studio\2017\Enterprise\VC\Auxiliary\Build\vcvarsall.bat"  (
+if EXIST "%ProgramFiles(x86)%\Microsoft Visual Studio\2017\Enterprise\VC\Auxiliary\Build\vcvarsall.bat"  (
```
KindDragon: "%ProgramFiles(x86)% or "%ProgramFiles% depends on your Windows (32-bit or 64-bit). Same VS140COMNTOOLS variable and others

src/engine/build.bat
```diff
@@ -106,19 +106,19 @@ if NOT "_%VS150COMNTOOLS%_" == "__" (
...
     set "BOOST_JAM_TOOLSET=vc15"
-    set "BOOST_JAM_TOOLSET_ROOT=%ProgramFiles%\Microsoft Visual Studio\2017\Enterprise\VC\"
+    set "BOOST_JAM_TOOLSET_ROOT=%ProgramFilesi(x86)%\Microsoft Visual Studio\2017\Enterprise\VC\"
     goto :eof)
 call :Clear_Error
-if EXIST "%ProgramFiles%\Microsoft Visual Studio\2017\Professional\VC\Auxiliary\Build\vcvarsall.bat"  (
+if EXIST "%ProgramFilesi(x86)%\Microsoft Visual Studio\2017\Professional\VC\Auxiliary\Build\vcvarsall.bat"  (
```
teeks99: That was a typo.

src/engine/build.bat
teeks99: I had thought that windows 32-bit (Vista and up) defined both %ProgramFiles% and %ProgramFiles(x86)% to be the same C:\Program Files\. Then on 64-bit systems, %ProgramFiles% points to C:\Program Files\ and %ProgramFiles(x86)% points to C:\Program Files (x86)\.  Is that not correct? I haven't had a 32-bit machine for a long time, so I can't easily look.

Regardless, I think searching for a specific path is a bad long-term solution. There isn't any guarantee that a user will install it there (it is selectable during install). If the VS15COMNTOOLS variable isn't reliable, then maybe we can query the registry or something to make it work? 

I really just submitted this pull request as an interm fix because it was breaking my regression tests on a machine that happened to have msvc-15 installed on it.

KindDragon: > Is that not correct? I haven't had a 32-bit machine for a long time, so I can't easily look.

I don't have it too, but from my memory, old 32-bit Windows had only %ProgramFiles%. But I think it fine for now.

> If the VS15COMNTOOLS variable isn't reliable, then maybe we can query the registry or something to make it work?

It is not easy to do for VS2017 from scripting languages http://stackoverflow.com/q/41106407/61505

src/engine/build.bat
KindDragon: Can you send PR to fix it?

teeks99: In #161

src/engine/build.bat
teeks99: Yuck, hopefully when the RTM is made available in the next couple days they'll have a better solution.

AndrewPardoe:
Just a quick note: I'm on the MSVC team and am trying to find an answer for this from the VS folks.

KindDragon:
@AndrewPardoe Thank you. If you find out something, please let me know in issue #157

AndrewPardoe:
It's ugly, but commenting now.

PR #167, opened 2017-02-20
Fix toolset version for VS2017 (should be 14.1)

KindDragon:
Related to #157

### DIFF
--- a/src/tools/msvc.jam
+++ b/src/tools/msvc.jam
@@ -907,9 +907,9 @@ local rule configure-really ( version ? : options * )
             # version from the path.
             # FIXME: We currently detect both Microsoft Visual Studio 9.0 and
             # 9.0express as 9.0 here.
-            if [ MATCH "(Microsoft Visual Studio 15)" : $(command) ]
+            if [ MATCH "(Microsoft Visual Studio[\/\\]2017)" : $(command) ]
             {
-                version = 15.0 ;
+                version = 14.1 ;
             }
             else if [ MATCH "(Microsoft Visual Studio 14)" : $(command) ]
             {
@@ -1595,7 +1595,7 @@ if [ MATCH (--debug-configuration) : [ modules.peek : ARGV ] ]
                      armv7 armv7s ;
 
 # Known toolset versions, in order of preference.
-.known-versions = 15.0 14.0 12.0 11.0 10.0 10.0express 9.0 9.0express 8.0 8.0express 7.1
+.known-versions = 14.1 14.0 12.0 11.0 10.0 10.0express 9.0 9.0express 8.0 8.0express 7.1
     7.1toolkit 7.0 6.0 ;
 
 # Version aliases.
@@ -1608,7 +1608,7 @@ if [ MATCH (--debug-configuration) : [ modules.peek : ARGV ] ]
 .version-alias-11 = 11.0 ;
 .version-alias-12 = 12.0 ;
 .version-alias-14 = 14.0 ;
-.version-alias-15 = 15.0 ;
+.version-alias-14.1 = 14.1 ;
 
 # Names of registry keys containing the Visual C++ installation path (relative
 # to "HKEY_LOCAL_MACHINE\SOFTWARE\\Microsoft").
@@ -1624,7 +1624,6 @@ if [ MATCH (--debug-configuration) : [ modules.peek : ARGV ] ]
 .version-11.0-reg = "VisualStudio\\11.0\\Setup\\VC" ;
 .version-12.0-reg = "VisualStudio\\12.0\\Setup\\VC" ;
 .version-14.0-reg = "VisualStudio\\14.0\\Setup\\VC" ;
-.version-15.0-reg = "VisualStudio\\15.0\\Setup\\VC" ;
 
 # Visual C++ Toolkit 2003 does not store its installation path in the registry.
 # The environment variable 'VCToolkitInstallDir' and the default installation
--- a/src/tools/msvc.py
+++ b/src/tools/msvc.py
@@ -773,7 +773,7 @@ def configure_really(version=None, options=[]):
             # FIXME: We currently detect both Microsoft Visual Studio 9.0 and
             # 9.0express as 9.0 here.
             if re.search("Microsoft Visual Studio[\/\\]2017", command):
-                version = '15.0'
+                version = '14.1'
             elif re.search("Microsoft Visual Studio 14", command):
                 version = '14.0'
             elif re.search("Microsoft Visual Studio 12", command):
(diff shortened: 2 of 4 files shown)


teeks99:
Just to be clear, the version of the c++ compiler is "15", it is the CRT that is only being incremented to "14.1".

https://build2.org/article/notes-visual-studio-15.xhtml

src/tools/msvc.jam
```diff
@@ -1624,7 +1624,7 @@ if [ MATCH (--debug-configuration) : [ modules.peek : ARGV ] ]
 .version-11.0-reg = "VisualStudio\\11.0\\Setup\\VC" ;
 .version-12.0-reg = "VisualStudio\\12.0\\Setup\\VC" ;
 .version-14.0-reg = "VisualStudio\\14.0\\Setup\\VC" ;
-.version-15.0-reg = "VisualStudio\\15.0\\Setup\\VC" ;
+.version-14.1-reg = "VisualStudio\\14.1\\Setup\\VC" ;
```
jhunold: This looks like classic copy&paste. I don't have any of those registry values. Any hints on the real registry values are welcome.

src/tools/msvc.py
```diff
@@ -1293,7 +1293,7 @@ def first(self):
 __version_11_0_reg = "VisualStudio\\11.0\\Setup\\VC"
 __version_12_0_reg = "VisualStudio\\12.0\\Setup\\VC"
 __version_14_0_reg = "VisualStudio\\14.0\\Setup\\VC"
-__version_15_0_reg = "VisualStudio\\15.0\\Setup\\VC"
+__version_14_1_reg = "VisualStudio\\14.1\\Setup\\VC"
```
jhunold: and copy&paster error, too.

src/tools/msvc.jam
KindDragon: They don't have register keys for VS2017. Should I just remove it?
How to support properly VS2017 path detection is  still being debated in #157

jhunold: Yes, it is better to remove them in this case. I've taken a look into #157 but I'm not sure which way to go. And I'm still not convinced that we call the thing "14.1"

KindDragon:
> I've taken a look into #157 but I'm not sure which way to go. And I'm still not convinced that we call the thing "14.1"

If it's similar to VS 2013 - 7.1, so it it should be 14.1. But for VS2013 IDE version was also 7.1 (IDE version for VS2017 15.0)

KindDragon:
```
ECHO ### Toolsets supported by this script are: borland, como, gcc, gcc-nocygwin,
ECHO ###     intel-win32, metrowerks, mingw, msvc, vc7, vc8, vc9, vc10, vc11, vc12, vc14
```
These versions match with ide version or toolset version (crt dll).
Cl.exe version is very different https://build2.org/article/notes-visual-studio-15.xhtml

teeks99:
The more I look at this, the more I wonder what microsoft was thinking...

The directory that the visual studio tools are installed in is `%PROGRAMFILES%\MicrosoftVisualStudio\2017\[Community|Professional|Enterprise]\VC\Tools\MSVC\14.10.24930\bin\HostX[86|64]\x[86|64]\cl.exe`
Note that is 14.10, not 14.1!

Looking at the properties of cl.exe, the "File version" is 19.10.24930.0, the "Product version" is 14.10.24930.0. 

Looking at the runtime, in `%PROGRAMFILES%\MicrosoftVisualStudio\2017\[Community|Professional|Enterprise]\VC\Redist\MSVC\14.10.24930\x[86|64]\Microsoft.VC150.CRT\` you have `concrt140.dll`, `msvcp140.dll`, `vccorlib140.dll`, `vcruntime140.dll`.  

So there is a 14.10 directory, that contains modules (side-by-side assemblies?) labeled with "VC150", containing DLLs that specify "140". I don't think the runtime is generally relevant to us, but this is going to cause major confusion among developers!

Looking at the versions in visual studio under help->about, the displayed info is:
    Microsoft Visual Studio Community 2017 RC
    VisualStudio/15.0.0-RC.4+26206.0

The actual tool lines such as "Visual C++ 2017 RC" only have some kind of UID next to them that doesn't seem to translate into a version number (just like in VS2015).

I vote we call the toolset "msvc-15.0" and "vc15", as microsoft is loudly trumpeting this as visual studio version "15".

teeks99:
At one point I think there was a link to a microsoft blog post explaining all this posted to the mailing list, does anyone happen to have a link to that? I'm not having much luck finding it.

teeks99:
I've also found [this page](https://www.visualstudio.com/en-us/productinfo/vs2017-install-product-Enterprise.workloads), updated today that lists the various build tools as being 15.0.X.

Specifically this line:
Microsoft.VisualStudio.Component.VC.Tools.x86.x64 	VC++ 2017 v141 toolset (x86,x64) 	15.0.26109.1 	Recommended

KindDragon:
> I've also found this page, updated today that lists the various build tools as being 15.0.X.

I think this comment related to your question https://www.reddit.com/r/cpp/comments/5ut2bw/visual_studio_2017_isnt_supported_by_boost/ddzgtcm/

> Microsoft.VisualStudio.Component.VC.Tools.x86.x64 VC++ 2017 v141 toolset (x86,x64) 15.0.26109.1

It's toolset installer for VS2017, so it should be 15.00. But yes it's confusing

KindDragon:
> Are we sure the VS top-line version is 14.1 for VS 2017?

Yes https://github.com/boostorg/config/pull/119#issuecomment-281500274

KindDragon:
But this is the version that would be in dll name, like `boost_filesystem-vc140-mt-1_6X.dll`.
VS2017 can use dll's from VS2015 or VS2017, so I think they should be named `boost_filesystem-vc140-mt-1_6X.dll` and `boost_filesystem-vc141-mt-1_6X.dll` 

And we can't change this version for old VS versions

KindDragon:
JFI Boost Library Naming http://www.boost.org/doc/libs/1_42_0/more/getting_started/unix-variants.html#library-naming

KindDragon:
IDK why Boost toolset enumeration contains version only for Visual Studio

MaartenBent:
Looking at the Platform Toolsets available for a project in Visual Studio 2017, It should indeed be 141:
vs-toolsets

refack:
Do we have a bottom line for me to improve #170?

KindDragon:
Quick voting. Add reaction to this post 👍 if you think toolset should be 14.10, 👎 if you think it should remain 15.00, or 😕 if you uncertain

refack:
Voting with my feet at #170

PR #170, opened 2017-02-28
fix #157: detect vs2017 (vc141) and setup environment

### LINKED ISSUE #157: Support VS2017

refack:
Based on JIT compilation of a C# file and COM

### DIFF
--- a/src/tools/msvc.jam
+++ b/src/tools/msvc.jam
@@ -708,6 +708,19 @@ local rule auto-detect-toolset-versions ( )
         # Get installation paths from the registry.
         for local i in $(.known-versions)
         {
+            if $(i) = 14.10
+            {
+                local file = [ path.make [ modules.binding $(__name__) ] ] ;
+                local cl141_path = [ path.native [ path.join [ path.parent $(file) ] vc141helper cl141_path.cmd ] ] ;
+                local shell_ret = [ SHELL $(cl141_path) ] ;
+                local cl_path = $(shell_ret:D) ;
+                if $(cl_path)
+                {
+                    path = [ path.native $(cl_path) ] ;
+                    register-configuration $(i) : $(path) ;
+                }
+            }
+
             if $(.version-$(i)-reg)
             {
                 local vc-path ;
@@ -820,7 +833,16 @@ local rule generate-setup-cmd ( version : command : parent : options * : cpu : g
         }
         else
         {
+            if [ MATCH "(14.10)" : $(version) ]
+            {
+                if $(.debug-configuration)
+                {
+                    ECHO 'notice: [generate-setup-cmd] $(version) is 14.10' ;
+                }
+                parent = [ path.native [ path.join  $(parent) "..\\..\\..\\..\\..\\Auxiliary\\Build" ] ] ;
+            }
             setup = [ locate-default-setup $(command) : $(parent) : $(default-setup) ] ;
+            setup ?= [ path.join  $(parent) "vcvarsall.bat" ] ;
         }
     }
     
@@ -894,6 +916,10 @@ local rule configure-really ( version ? : options * )
 
         local command = [ feature.get-values <command> : $(options) ] ;
 
+        # For 14.10 we need the exact version as MS is planning rolling updates
+        # that will cause our `setup-cmd` to become invalid
+        exact-version = [ MATCH "(14\.10\.[0-9\.]+)" : $(command) ] ;
+
         # If version is specified, we try to search first in default paths, and
         # only then in PATH.
         command = [ common.get-invocation-command msvc : cl.exe : $(command) :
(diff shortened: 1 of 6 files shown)


src/tools/msvc.jam
```diff
@@ -907,7 +925,7 @@ local rule configure-really ( version ? : options * )
             # version from the path.
             # FIXME: We currently detect both Microsoft Visual Studio 9.0 and
             # 9.0express as 9.0 here.
-            if [ MATCH "(Microsoft Visual Studio 15)" : $(command) ]
+            if [ MATCH "(14.10)" : $(command) ]
```
KindDragon: This line should be `if [ MATCH "(Microsoft Visual Studio[\/\\]2017)" : $(command) ]`

KindDragon:
Maybe transfer functionality for debugging in a separate PR to speed up the merger of this PR?

src/tools/msvc.jam
refack: I know but with 2017 that string is not a part of the path. Only predictable string is `14.10`

refack: On my machine it's `C:\bin\dev\VS\2017\BuildTools\VC\Tools\MSVC\14.10.24930\bin\HostX64\x64`

refack:
@KindDragon what debugging functionality specifically?

src/tools/msvc.jam
KindDragon: I think this codepath only for default path to VS, check expression for VS2015 below for example

KindDragon:
Ahh, your PR based on changes from debug. I think you should remove this commits from your PR

src/tools/msvc.jam
refack: It's logic to deduce the VS version from a given path. `14.10` is the only stable path part (`MSVC\14.10` at most).
There are other patterns in this block RE: [line 965](https://github.com/refack/build/blob/4973c19e47efbe338664145818fdc00fb24759e9/src/tools/msvc.jam#L965)
VS2017 setup has been reworked extensively...

refack:
Branch base oops (`debug` and `develop` look too much alike for a dysxelic ;)

refack:
👍 14.10

refack:
@grafikrobot actually it has nuance, The "whole" Visual Studio suite will be version 15.0, but the C++ toolset will be 14.10

refack:
For this project `15.0` is used only as an internal moniker, so can consider it a nuance.... 🤷
RE: dotnet/core#509

KindDragon:
@refack Discussion about 14.10 in PR https://github.com/boostorg/build/pull/167

src/tools/vc141helper/GetVS2017Configuration.cs
```diff
@@ -0,0 +1,272 @@
+﻿// powershell -ExecutionPolicy Unrestricted -Version "2.0" -Command "&{ Add-Type -Path Program.cs; [VisualStudioConfiguration.Program]::Main(@())}"
```
grafikrobot: This needs a license / copyright statement to be considered for inclusion in Boost.

refack: done

refack:
@grafikrobot checking

refack:
@grafikrobot can you gist the output of `try_powershell.cmd` just to make sure you have the VCTools "workload" installed.
[Here's mine](https://gist.github.com/refack/a0868d2d2c3865dd3b2c8aaca0d361b4#file-output-josn-L80)

refack:
@grafikrobot I tried to simplify the non C# part, and added some fallbacks.

refack:
[CI machine](https://ci.appveyor.com/project/refack/windows-autoconf/build/job/tksj2vuno0aajr1o) passes

jhunold:
Tested the last version of the patch with 2017 Enterprise. Works

refack:
One small step for me, kinda small step for `Boost.build` 🐱‍🚀

AndrewPardoe:
The officially supported vswhere tool is now available as part of the install starting today with Visual Studio 15.2 preview 2: https://blogs.msdn.microsoft.com/heaths/2017/04/21/vswhere-is-now-installed-with-visual-studio-2017/. This doesn't fix all the existing installs of VS 2017 but going forward you can now detect the VS install locations and whether C++ tools are available from a scripted environment. I hope this helps your scenario.

PR #171, opened 2017-03-09
Added python tool examples

teeks99:
This is an addition to the documentation that includes how to configure the python tool. I think I've put it in correctly, but I don't have the toolchain to actually build the docs and see what they look like. Can someone who does have the toolchain try that before this is merged?

### DIFF
--- a/doc/src/reference.xml
+++ b/doc/src/reference.xml
@@ -1774,6 +1774,121 @@ using bzip2 : 1.0.6 : &lt;source&gt;C:/Devel/src/bzip2-1.0.6 : &lt;toolset&gt;ms
 using bzip2 : 1.0.6 : : &lt;toolset&gt;gcc ;
 </programlisting>
         </section>
+
+        <section id="bbv2.reference.tools.libraries.python">
+          <title>Python</title>
+          <indexterm><primary>Python</primary></indexterm>
+
+          <para>Provides support for the
+          <ulink url="http://www.python.org">python</ulink> language
+          environment to be linked in as a library.</para>
+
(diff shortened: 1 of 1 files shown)


doc/src/reference.xml
```diff
@@ -1783,6 +1783,121 @@ using bzip2 : 1.0.6 : &lt;source&gt;C:/Devel/src/bzip2-1.0.6 : &lt;toolset&gt;ms
...
+: # conditions
+;
+
+# On windows
+using python
+: 2.7 # version
+: C:\\Python27-32\\python.exe # Interperter/path to dir
+: C:\\Python27-32\\include # includes
+: C:\\Python27-32\\libs # libs
+: <address-model>32 <address-model> # conditions - both 32 and unspecified
```
grafikrobot: I think those "<>" need to be encoded to "&lt;", "&gt;".

doc/src/reference.xml
```diff
@@ -1783,6 +1783,121 @@ using bzip2 : 1.0.6 : &lt;source&gt;C:/Devel/src/bzip2-1.0.6 : &lt;toolset&gt;ms
...
+: C:\\Python27-32\\libs # libs
+: <address-model>32 <address-model> # conditions - both 32 and unspecified 
+;
+
+using python
+: 2.7 # version
+: C:\\Python27-64\\python.exe # Interperter/path to dir
+: C:\\Python27-64\\include # includes
+: C:\\Python27-64\\libs # libs
+: <address-model>64 # conditions
```
grafikrobot: "<>" need to be encoded.

PR #175, opened 2017-03-13
Documentation Updated with Visual Studio 2017 version and example

teeks99:
This updates the toolset documentation to include the visual studio 2017 version as well as
a basic msvc usage example. 

I think I've put it in correctly, but I don't have the toolchain to actually build the docs and see what they look like. Can someone who does have the toolchain try that before this is merged?

### DIFF
--- a/doc/src/reference.xml
+++ b/doc/src/reference.xml
@@ -1021,6 +1021,7 @@ using gcc : &toolset_ops; ;</programlisting>
           C++</ulink> command-line tools on Microsoft Windows. The supported
           products and versions of command line tools are listed below:</para>
           <itemizedlist>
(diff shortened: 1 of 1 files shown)


doc/src/reference.xml
```diff
@@ -1032,6 +1033,14 @@ using gcc : &toolset_ops; ;</programlisting>
             <listitem><para>Visual Studio 6.0, Service Pack 5&#x2014;6.5</para></listitem>
           </itemizedlist>
 
+          <para>The user would then call the boost build executable with the
+          toolset set equal to <command>msvc-[version number]</command> for 
+          example to build with Visual Studio 2017 one could run:
+          <programlisting>
+.\b2 toolset=msvc-14.1 target
```
KindDragon: I think it should be msvc-14.10

PR #183, opened 2017-03-19
Updated all the references to 14.10 and 1410 to 14.1 and 141

teeks99:
This PR should *not* be accepted unless the [community consensus on the mailing list](http://lists.boost.org/Archives/boost/2017/03/233543.php) converges around option 2.


teeks99:
I think I found all the places that the changes got moved to, hopefully nothing else got trashed in the merge.

PR #184, opened 2017-03-19
build.bat sanity

refack:
A good effort to kill cases of endless recursion


KindDragon:
I think it can be closed now

PR #190, opened 2017-04-02
ci: Enable DO_DIFF on travis-ci to facilitate debugging.

tee3:
The Boost.Build tests provide a mechanism to observe the difference between the expected output of a test and the output of a failing test.  This is useful in a CI context because it is time-consuming to resolve issues that occur only on the cloud-based CI system.

I don't believe that this will cause issues due to large prints as most tests succeed and print nothing.  It is only when a test fails does this print anything and this should be rare for CI.  Furthermore, I believe that most tests do not typically print out large chunks of text when running.

This has passed the testing on travis-ci on a local branch.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,9 +2,9 @@ sudo: false
 os:
   - linux
 env:
-  - TOOLSET=gcc TEST_ALL_EXTRAS= DO_DIFF=1
+  - TOOLSET=gcc TEST_ALL_EXTRAS=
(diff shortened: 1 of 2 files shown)


tee3:
Sure thing.

tee3:
I've added it to appveyor.  I tested it separately by echoing the value of %DO_DIFF% and it appears to work properly.

Note that the appveyor builds are all failing for other reasons, I believe to the ongoing development of the VC2017 toolset.

PR #195, opened 2017-04-25
Add a feature for using quickbook's strict mode

danieljames:
I've just added strict mode to develop in the quickbook module.


danieljames:
I've already reverted it:

https://github.com/boostorg/quickbook/commit/0223ebb9385452699b335887e4e510843e4204f8

PR #198, opened 2017-05-18
Replace pgCC with pgc++

Rombur:
`pgCC` does not exist more. The last version of PGI that supported it is 15.10.


Rombur:
@grafikrobot pgi 15.10 was released in October 2015. I found that `pgc++` was introduced in version 13.1 (January 2013) for linux. I don't know when it was introduced on other platforms.

PR #199, opened 2017-05-22
Fix build errors with stdlib=sun-stlport on Linux 

mkartashev:
For sun-stlport, add -compat=5 to the command line in order to avoid build 
errors on the platforms where it is not the default (like on Linux). Without
this additional option, the compiler exits with an error because
-library=stlport4 only works with the Sun ABI (-compat=5), while the default
might be GNU ABI (-compat=g).

Do the same for the stdcxx library that is also available only in -compat=5
mode. This change is not strictly necessary as the stdcxx library is only
available on Solaris, where the default is -compat=5 anyway. It has been added
as a precaution.


mkartashev:
Corresponding ticket is https://svn.boost.org/trac/boost/ticket/13042

PR #208, opened 2017-07-03
Add lzma.jam file.

rdoeffinger:
Just a quick hack to allow to implement LZMA
compressor and decompressor in iostreams (see also
trac ticket 7534).
Probably needs to be refined at some point by someone
who knows what they are doing, but it seems to work in principle.


rdoeffinger:
Any comments? I know it's not ideal to just add a (currently) unused file, but it seems the simplest approach.
The iostreams change depending on this: https://github.com/boostorg/iostreams/pull/38

PR #209, opened 2017-07-07
Add support for <arch-and-model> in common.format-name

pdimov:


src/tools/common.jam
```diff
@@ -1029,6 +1033,19 @@ local rule address-model-tag ( name : type ? : property-set )
     return [ $(property-set).get <address-model> ] ;
 }
 
+# Create a tag for the architecture and model
+# <architecture>x86 <address-model>32 would generate "x32"
```
stefanseefeld: So windows-on-intel would generate "x32" and "x64". What about ARM, PPC, etc. ? Do they generate "A32", "P64", etc. ? I'd rather Boost respects preexisting nomenclature for naming architectures, including "x86", "x86_64", "arm", "aarch64", etc.

pdimov: Yes, they generate `a32`, `p64`, and so on.

pdimov:
It would not be preferable to me. The current scheme reflects the underlying Boost.Build properties <architecture> and <address-model>, including the fact that they are independent. I do not view their obfuscation by replacement with well-established, but arbitrary and irregular, names, as an improvement.

PR #219, opened 2017-07-27
Add block comments to jam language.

grafikrobot:
This adds non-nested block comments wrapped with "#*" and "*#". This
should help in cleaning up embedded documentation and generally make
writing jam code easier.

### DIFF
--- a/doc/bjam.qbk
+++ b/doc/bjam.qbk
@@ -393,7 +393,7 @@ The arguments starting with the "=--option=" forms are passed to the =build.jam=
 must be quoted with double quotes (") to be used as arbitrary tokens, such as
 variable or target names.
 
-Comments start with the [^#] character and extend until the end of line.
+Comments start with the [^#] character and extend until the end of line. And block comments start with [^#*] and extend until the next [^*#].
 
 [endsect]
 
--- a/doc/bjam.qbk
+++ b/doc/bjam.qbk
@@ -393,7 +393,7 @@ The arguments starting with the "=--option=" forms are passed to the =build.jam=
 must be quoted with double quotes (") to be used as arbitrary tokens, such as
 variable or target names.
 
-Comments start with the [^#] character and extend until the end of line. And block comments start with [^#*] and extend until the next [^*#].
+Comments start with the [^#] character and extend until the end of line. And block comments start with [^#|] and extend until the next [^|#].
 
 [endsect]
 
--- a/example/time/jamroot.jam
+++ b/example/time/jamroot.jam
@@ -1,6 +1,8 @@
-# Distributed under the Boost Software License, Version 1.0. (See
-# accompanying file LICENSE_1_0.txt or copy at
-# http://www.boost.org/LICENSE_1_0.txt)
+#*
+Distributed under the Boost Software License, Version 1.0. (See
+accompanying file LICENSE_1_0.txt or copy at
+http://www.boost.org/LICENSE_1_0.txt)
+*#
 
 #[jamroot
 #<< Import the time rule from the testing module.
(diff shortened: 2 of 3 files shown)


swatanabe:
AMDG

On 07/27/2017 07:56 AM, Rene Rivera wrote:
> This adds non-nested block comments wrapped with "#*" and "*#". This
> should help in cleaning up embedded documentation and generally make
> writing jam code easier.

This seems somewhat dangerous:

# * ... - this is common plain text formatting of a list
#* ... - this is the beginning of a block comment

In Christ,
Steven Watanabe

frenchtoast747:
Would it be too difficult to incorporate Python's docstring as an entity?
```
local foo = bar ;
"""
this is foo. it does bar.
"""
```

I'm assuming that in order to do that, though, it would require a semicolon and actually become a statement. And/or it would make the grammar too difficult to deal with.

Side note, I have used action declarations as block comments:
actions {
  hello there! This is just a raw string, so anything goes.
}

But, I typically only do that in order to comment out chunks of code, rather than as comments.

swatanabe:
AMDG

On 07/27/2017 08:20 AM, Rene Rivera wrote:
> @swatanabe Good point.. Do you have a suggestion for another style to use? Some possibilities I can think of are (in order or preference) `#{ }#`, `#| |#`, `#/ /#`, `#: :#`, `#~ ~#`.
> 

  `#{` is a bad idea because it interferes with commented out code.
`#:` has the same problem, though less so, since ':' rarely appears
at the beginning of a line.  For `#~`, I think I've seen a number
of .jam files with comments formatted like this.  I don't know why,
however.  Is there any precedent in other languages that use `#`
for comments?
  Actually, I'm not really sure why you want this is the first place.
I've never found the regular # comments to be particularly annoying.
The main place where you need block comments is when documenting
parameters inside the parameter list, but if you're putting comments
there, you'll need line breaks anyway or the lines will get way
too long.

In Christ,
Steven Watanabe

swatanabe:
AMDG

On 07/27/2017 07:54 PM, Rene Rivera wrote:
> You are close.. The use case I have in mind is for adding embedded documentation. In particular for tools like quickbook and acsiidoctor that support including documentation segments from "arbitrary" files. Even though we have the embedded "--help" system, it's cumbersome and it's rarely used. As people prefer to read documentation in HTML or PDF form instead. I'm trying to add the embedded documentation similar to how I document the Predef library. It's much easier to keep such documentation up to date and be more comprehensive. Without the block comments the tools need to have special handling for the commenting format. And I really want to avoid more custom tool work.
> 

  In theory, the existing help system can generate
html, although I haven't actually tried it.  Also,
# is one of the most widely used comment formats.
I'm not convinced that creating a new unique form
of comment is a good idea.  You can always preprocess
the files with grep '^#`' (to match quickbook's //`)
or some such.

In Christ,
Steven Watanabe

swatanabe:
AMDG

On 07/27/2017 09:13 PM, Rene Rivera wrote:
> 
> Lets see comment styles with `#`..
> - Algol 68: `# BlockComment #`
> - Various shells: `<# BlockComment #>`
> - Curl: `|# BlockComment #|`
> - Cobra: `/# BlockComment #/`
> - Lisp/Scheme: `#| BlockComment |#`
> - Perl: #`{ BlockComment }`
> - Python: `''' BlockComment '''` `""" BlockComment """`

Well, technically this is a string...

> - Seed7: `(* BlockComment *)`
> 
> So maybe the Lisp/Scheme style?

+1.

> We could probably also do `<# #>`. But not starting with `#` seems risky in collisions with use inside regular tokens.
> 

  Yeah.  It wouldn't matter if we had a smarter tokenizer,
but with the way things are, `#` can appear in the middle
of a token.

In Christ,
Steven Watanabe

PR #225, opened 2017-08-09
Fix for #13154: added -features=zla to the Oracle compiler's command line

mkartashev:
Fix for #13154: added -features=zla to the compiler's command line for both libraries that require -compat=5 (stlport and apache). For the GNU library, -compat=g is used through the -std=c++03 option and it includes -features=zla


mkartashev:
Thanks!

PR #228, opened 2017-08-16
VxWorks support for b2 

kuhlenough:
While the source contains VxWorks specific defines and configuration, the process of the building in a cross-build environment has been hackneyed. 
For example:  
https://svn.boost.org/trac10/wiki/Guidelines/VxWorks

This is less than ideal, as VxWorks of any vintage supports at minimum 3 processor families and utilizes multiple compilers that vary by specific processor. So any given developer must adapt any instructions to there environment.   

This pull adds the ability to use b2 with VxWorks, and and more importantly begins the path to having each VxWorks developer being able to run test suite against there particular BSP (board support package.)

Additional pulls  to add supporting changes to config, test, and individual libraries will follow.
Current test results, supporting configuration examples, and instructions found here:
https://github.com/Wind-River/vxworks7-layer-for-boost


kuhlenough:
I see this pull now has conflict.  
Is there a reason it has not been merged previously?
Would you like the changes in another format?

kuhlenough:
fixed conflicts

PR #235, opened 2017-09-03
- fix "rm is not recognized" error when compiling with Clang on Windows

SSE4:


src/tools/clang-linux.jam
```diff
@@ -146,19 +147,28 @@ actions compile.c.with-pch bind PCH_FILE
...
 
+if [ os.name ] = NT
+{
+    RM = del /f /q ;
+}
+else
+{
+    RM = rm -f ;
+}
+
```
grafikrobot: There's already an existing definition of RM that you can use. Just do:
```
import common ;
RM = [ common.rm-command ] ;
```

SSE4: done, thanks for pointing

PR #244, opened 2017-09-29
fix typos

hia3:
I hope "ass needed" is not another idiom/abbreviation like "iff" for example.

### DIFF
--- a/doc/src/faq.xml
+++ b/doc/src/faq.xml
@@ -433,8 +433,8 @@ alias my-lib
       automatically added to all of its dependants' build properties. The
       dependants need not care if <code>my-lib</code> is a header-only or not,
       and it is possible to later make <code>my-lib</code> into a regular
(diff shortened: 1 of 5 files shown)


doc/src/faq.xml
```diff
@@ -434,6 +434,7 @@ alias my-lib
       dependants need not care if <code>my-lib</code> is a header-only or not,
       and it is possible to later make <code>my-lib</code> into a regular
       compiled library without having to that its dependants' declarations.
+      <!-- Even if "without having to that its" is correct English, I'd suggest to simplify this sentence. -->
```
eldiener: I think the correction should be:
"without having to add that to its dependants' declarations."
Even in this correction the word 'that' is still a bit ambiguous, so maybe a better correction would be:
"without having to add the includes to its dependants' declarations."

PR #250, opened 2017-10-05
Use python-config to get python paths

piwaniuk:
Sometimes python appends additional letters after a version number in its paths (for example `/usr/inlude/python3.6m`). This caused an error when building `boost_python` with default configuration, as described on [https://svn.boost.org/trac10/ticket/11120](https://svn.boost.org/trac10/ticket/11120). The patch enables `python-config` command to determine library and include paths.

### DIFF
--- a/src/tools/python.jam
+++ b/src/tools/python.jam
@@ -503,6 +503,22 @@ local rule probe ( python-cmd )
 }
 
 
+# Use python-config command to get default paths
+#
+local rule compute-python-config-paths ( version ? : exec-prefix ? )
+{
+    local cfg-cmd = $(exec-prefix)/bin/python$(version)-config ;
+
+    local includes-cmd =
+        $(cfg-cmd)" --includes | sed -n 's/^-I\(.*\) .*/\1/p' | tr -d '\\n'" ;
+    includes ?= [ shell-cmd $(includes-cmd) ] ;
+
+    local libraries-cmd =
+	      $(cfg-cmd)" --ldflags | sed -n 's/^-L\(.*\) .*/\1/p' | tr -d '\\n'" ;
+    libraries ?= [ shell-cmd $(libraries-cmd) ] ;
+}
+
+
 # Make sure the "libraries" and "includes" variables (in an enclosing scope)
 # have a value based on the information given.
 #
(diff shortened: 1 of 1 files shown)


tkelman:
There seem to be some python 3 environments that have a `python-config` but not a `python3.6-config` executable. I was able to use the patch from https://svn.boost.org/trac10/attachment/ticket/11120/python_jam.patch, though `sys.abiflags` isn't available in python 2 so it throws a nuisance warning.

HDembinski:
As this is a long-standing issue which dates back three years
https://svn.boost.org/trac10/ticket/11120#comment:21
It is great to see an attempt to solve this. Since there are issues with the solution based on `python-config`, as pointed out by @tkelman , let me propose another solution.
According to my research, the safest way of getting python paths is via python itself. The following two code snippets print the python include path and the python library, by pulling the information out of distutils.
```
# get python include path
from distutils import sysconfig
import sys
sys.stdout.write(sysconfig.get_python_inc())
```

```
# get python library
from distutils import sysconfig
import os.path as op
import sys
v = sysconfig.get_config_vars()
fpaths = [op.join(v[pv], v['LDLIBRARY']) for pv in ('LIBDIR', 'LIBPL')]
sys.stdout.write(list(filter(op.exists, fpaths))[0])
```

I use these scripts in https://github.com/hdembinski/histogram which compiles successfully on Travis with Python 2.7 and 3.6.

tkelman:
python-config itself isn't that much more complicated (https://github.com/pypa/virtualenv/blob/master/virtualenv_embedded/python-config) - the hard part can be what name it goes by. The built-in version of `python3-config` looks slightly different, but virtualenv may be making do with a single source across all python versions?

HDembinski:
The output of the Python snippets requires no parsing, though, while there is a good deal of sed regex used in this patch to retrieve the info from python-config. The Python solution is easier, IMHO.

tkelman:
Sure, implementing this by asking sysconfig isn't a bad solution, but if you're going to use that approach I'd recommend being consistent with exactly what python-config would return. That's much better tested across various different python build configurations, distros etc than returning the first library that exists.

HDembinski:
Fair point. So perhaps the maintainers could comment? Should I submit a separate PR addressing the same issue?

jefftrull:
Please do merge this.... or some similar solution.

marcovzla:
Please merge this or solve this issue somehow

filips123:
What is current state of this? It seems the issue is still present on Boost Build 1.73.0, tested on GitHub Actions with macOS and Python 3.5, but most likely also on other systems. Include path is `/Library/Frameworks/Python.framework/Versions/3.5/include/python3.5` while it should be `/Library/Frameworks/Python.framework/Versions/3.5/include/python3.5m`.

Although include path in Python 3.8 is again `python3.8` without `m`, this still prevents me from building on older versions.

PR #252, opened 2017-10-09
Prevent vcvarsall.bat from changing current directory

hia3:
Building with VS2017 15.3.5 produces following error:

c1: fatal error C1083: Cannot open source file: 'yyacc.c': No such file or directory

This happens because vsdevcmd_end.bat changes current directory to %USERPROFILE%\Source if it exists.

### DIFF
--- a/src/engine/config_toolset.bat
+++ b/src/engine/config_toolset.bat
@@ -172,7 +172,9 @@ if "_%BOOST_JAM_TOOLSET_ROOT%_" == "__" (
 if "_%BOOST_JAM_ARCH%_" == "__" set BOOST_JAM_ARCH=x86
 set BOOST_JAM_ARGS=%BOOST_JAM_ARGS% %BOOST_JAM_ARCH%
 
+pushd %CD%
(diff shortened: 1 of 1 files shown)


hia3:
Here is what vsdevcmd_end.bat does:

    @REM Set the current directory that users will be set after the script completes
    @REM in the following order:
    @REM 1. [VSCMD_START_DIR] will be used if specified in the user environment
    @REM 2. [USERPROFILE]\source if it exists
    @REM 3. current directory
    if "%VSCMD_START_DIR%" NEQ "" (
        cd /d "%VSCMD_START_DIR%"
    ) else (
        if EXIST "%USERPROFILE%\Source" (
            cd /d "%USERPROFILE%\Source"
        )
    )

pdimov:
Why has this been closed?

pdimov:
This problem has just been reported against 1.66 beta 1 RC 1: https://lists.boost.org/Archives/boost/2017/11/239824.php

hia3:
> Why has this been closed?
I wasn't sure why this patch works. Feel free to apply it.

hia3:
Simple `pushd .` is enough, no need for %CD%

pdimov:
Yes, I was going to say the same thing.

jhunold:
Created an edited version on #263. Will merge to develop as soon as CI passes.

PR #268, opened 2017-11-28
Add test/Jamfile

pdimov:
This adds a `Jamfile` in `test` that just calls `python test_all.py`, so that the `b2 submodule-dir/test` command invoked by the superproject Travis runs the tests.

### DIFF
--- a/test/Jamfile
+++ b/test/Jamfile
@@ -1,0 +1,6 @@
+import testing ;
+import type ;
+
+type.register PY : py : EXE ;
+
+run test_all.py : : : <testing.launcher>python ;


swatanabe:
This doesn't work for a few reasons:
- ``using python ;`` in user-config.jam will break it because of duplicate registration of the PY type.
- ``b2 path/to/boost-build/test`` fails because the tests assume that the current working directory is test/ in several places.
- There are no explicit dependencies and no include scanner is set up, so the tests will only be rebuilt when test_all.py is modified.

swatanabe:
I used run-pyd instead of run.  This requires that python can be fully configured (including finding libraries and headers), but that's more a problem with the way python.jam is set up.  In the long run, I think letting python.jam handle running python scripts is the right way to go.  We have the facilities to do more fine-grained configuration checks, so python.jam should be adjusted to allow only finding the python executable at some point.

PR #278, opened 2018-01-12
configure VxWorks/Dinkum STL to match <rtti>/<exception-handling>

kuhlenough:


kuhlenough:
My goal is to have server contribute VxWorks build results to http://www.boost.org/development/tests/master/developer/summary.html
In my current involved VxWorks environment none of the <rtti> and <exception-handling> tests passed till I made this change.

PR #290, opened 2018-02-07
Add 'version-suffix' rule.

stefanseefeld:
I'd like to use the newly defined rule in the top-level Jamroot, to refine the `python-tag` rule to include the Python version suffix.


stefanseefeld:
It seems all the test failures in Travis CI are unrelated. I'd appreciate if someone could quickly review and hopefully merge this PR, so I can work on the code that use that new function.
Thanks,

PR #291, opened 2018-02-24
Enable detection of python's build flags for include dir

gnaggnoyil:
Many Linux distributions would put python3's include header into directories like `/usr/include/python3.6m` instead of  paths like `/usr/include/python3.6` where `m` is python's abi build flag. Previously the automatic guess for python's header include directory did not consider such situation, thus the commit. The implementation itself is inspired by Python's distutil module(or more precisely, `distuils.sysconfig.get_python_inc` function).

### DIFF
--- a/src/tools/python.jam
+++ b/src/tools/python.jam
@@ -545,6 +545,12 @@ local rule compute-default-paths ( target-os : version ? : prefix ? :
     else
     {
         includes ?= $(prefix)/include/python$(version) ;
+        if $(sys.abiflags)
+        {
+            # Calling "sys.abiflags" does not throw an exception, which 
+            # indicates this is not a build from source distro.
+            includes ?= $(includes)$(sys.abiflags) ;
+        }
 
         local lib = $(exec-prefix)/lib ;
         libraries ?= $(lib)/python$(version)/config $(lib) ;
@@ -746,7 +752,7 @@ local rule configure ( version ? : cmd-or-prefix ? : includes * : libraries ? :
     {
         # Values to be extracted from python's sys module. These will be set by
         # the probe rule, above, using Jam's dynamic scoping.
-        local sys-elements = version platform prefix exec_prefix executable ;
+        local sys-elements = version platform prefix exec_prefix executable abiflags ;
         local sys.$(sys-elements) ;
 
         # Compute the string Python's sys.platform needs to match. If not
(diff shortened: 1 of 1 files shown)


src/tools/python.jam
```diff
@@ -545,6 +545,13 @@ local rule compute-default-paths ( target-os : version ? : prefix ? :
     else
     {
         includes ?= $(prefix)/include/python$(version) ;
+        if $(sys.abiflags)
+        {
+            # Calling "sys.abiflags" does not throw an exception, which 
+            # indicates this is neither a Python 2 build, nor a build from
+            # source distro.
+            includes ?= $(includes)$(sys.abiflags) ;
```
tkelman: I think this should be `=` rather than `?=` otherwise it won't change anything.

tkelman:
It doesn't seem to immediately be breaking anything, but I'll note that on python 2.7, applying this patch now gives
```
warning: No python installation configured and autoconfiguration
note: failed.  See http://www.boost.org/libs/python/doc/building.html
note: for configuration instructions or pass --without-python to
note: suppress this message and silently skip all Boost.Python targets
```
likely because `sys.abiflags` doesn't exist there?

src/tools/python.jam
gnaggnoyil: How do `=` and `?=` differ? I'm not very familiar with boost.build and I thought `?=` would skip assignment if variable is already set and `=` would force an assignment.

gnaggnoyil:
@tkelman Yes Python 2 does not have `sys.abiflags` nor uses abiflags to form their include directory. But as `local rule probe` indicates when configuring for Python 2 `sys.abiflags` is just void thus `includes` should be left unchanged. I feel a bit strange that such an error would occur. I can confirm such patch works as intended when building for Python 3, I'll see if this patch really has problems for Python 2.

src/tools/python.jam
tkelman: I'm not that familiar either but when I was trying this patch locally it didn't solve the problem as-is. The setting of `includes` a few lines above is unconditional so it seems this line with a `?=` is resulting in a no-op.

tkelman:
We could possibly consider using `distutils.sysconfig` (or `sysconfig` for 2.7 and later) to get the include paths directly, which is the approach used by `python-config`. The assumption that include paths are directly under `sys.prefix` turns out to not be true when using `venv`.

PR #294, opened 2018-04-03
Avoid 100% CPU utilization if no timeout is set

neuschaefer:
While building boost, I noticed that jam0 and bjam used 100% of a CPU.
Strace showed that they were calling poll with a zero timeout in a loop.
This is because:
- the logic in exec_wait() initializes select_timeout to globs.timeout
- globs.timeout is zero when no action timeout is specified
- poll interprets a zero timeout as "return immediately" rather than
  "wait indefinitely".

Fix this by using a poll timeout of one minute when globs.timeout is
zero. Note that the value of one minute is arbitrary: Other values
should work just as well.

### DIFF
--- a/src/engine/execunix.c
+++ b/src/engine/execunix.c
@@ -499,11 +499,17 @@ void exec_wait()
         {
             /* disable child termination signals while in select */
             int ret;
+            int timeout;
             sigset_t sigmask;
             sigemptyset(&sigmask);
             sigaddset(&sigmask, SIGCHLD);
             sigprocmask(SIG_BLOCK, &sigmask, NULL);
-            while ( ( ret = poll( wait_fds, WAIT_FDS_SIZE, select_timeout * 1000 ) ) == -1 )
+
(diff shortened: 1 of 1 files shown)


swatanabe:
AMDG

On 04/03/2018 09:36 AM, Jonathan Neuschäfer wrote:
> While building boost, I noticed that jam0 and bjam used 100% of a CPU.
> Strace showed that they were calling poll with a zero timeout in a loop.
> This is because:
> - the logic in exec_wait() initializes select_timeout to globs.timeout
> - globs.timeout is zero when no action timeout is specified
> - poll interprets a zero timeout as "return immediately" rather than
>   "wait indefinitely".
> 
> Fix this by using a poll timeout of one minute when globs.timeout is
> zero. Note that the value of one minute is arbitrary: Other values
> should work just as well.

  It's better to set the timeout to -1, which does mean
"wait indefinitely".  This matches the original behavior
of using select with a NULL timeout.

In Christ,
Steven Watanabe

neuschaefer:
Good idea, I'll update my patch.

PR #309, opened 2018-05-16
subfeatures are expanded by feature.add-defaults

### LINKED ISSUE #308: Build fails when gcc is impllicitly selected as the toolset
Running plain `b2` fails with this error:
```
gcc.compile.c++ bin/gcc/debug/foo.o
/bin/sh: 2: -fPIC: not found

       -fPIC -O0 -fno-inline -Wall -g     -c -o "bin/gcc/debug/foo.o" "foo.cpp"

...failed gcc.compile.c++ bin/gcc/debug/foo.o...
```

Running `b2 toolset=gcc` works as expected.

My `user-config.jam` only contains `using gcc ;`

git bisect pointed at commit 15c876025afb3caa2961642179b112f0532dcdf1, and after reverting different changes I found out that rule `feature.add-default` stopped expanding `<toolset>gcc` (it looks like if I manually specify the toolset to use, it gets expanded elsewhere).

grisumbras:
Fixes #308.

### DIFF
--- a/src/build/feature.jam
+++ b/src/build/feature.jam
@@ -1080,6 +1080,15 @@ rule add-defaults ( properties * )
     local missing-top = [ set.difference $(.all-top-features) : $(properties:G) ] ;
     local more =  [ defaults $(missing-top) ] ;
 
+    local xproperties = [ MATCH "^([^:]+)$" : $(properties) ] $(more) ;
+    for local p in $(xproperties)
+    {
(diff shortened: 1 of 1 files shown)


swatanabe:
AMDG

  This patch is not correct.  Subfeature defaults are
supposed to be handled in the while loop below.

In Christ,
Steven Watanabe

grisumbras:
Yes, I realized it later, I am writing a different patch.

PR #310, opened 2018-05-16
subfeatures are expanded by feature.add-defaults

### LINKED ISSUE #308: Build fails when gcc is impllicitly selected as the toolset

grisumbras:
Fixes #308 .

### DIFF
--- a/src/build/feature.jam
+++ b/src/build/feature.jam
@@ -1092,9 +1092,9 @@ rule add-defaults ( properties * )
         }
     }
 
-    local worklist = $(properties) ;
+    local worklist = $(properties) $(more) ;
     local expanded-from-composite ;
-    local to-expand = $(more) ;
+    local to-expand ;
     while $(worklist)
     {
         # Add defaults for subfeatures of features which are present.


swatanabe:
AMDG

On 05/16/2018 02:09 PM, Dmitry wrote:
> Fixes #308 .
> You can view, comment on, or merge this pull request online at:
> 
>   https://github.com/boostorg/build/pull/310
> 
> -- Commit Summary --
> 
>   * subfeatures are expanded by feature.add-defaults
> 
> -- File Changes --
> 
>     M src/build/feature.jam (4)
> 
> -- Patch Links --
> 
> https://github.com/boostorg/build/pull/310.patch
> https://github.com/boostorg/build/pull/310.diff
> 

Okay, I see what's happening now, but this patch
is still not quite right.  It breaks the case of
applying the default of a composite feature.  I
think $(more) needs to be in *both* worklist and
to-expand.

In Christ,
Steven Watanabe

grisumbras:
Ok, I updated the patch.

PR #324, opened 2018-07-09
Make `toolset=clang-win` initialize the clang-win toolset

pdimov:
This is more an RFC although it can be applied as-is; it makes `toolset=clang-win` initialize the `clang-win` toolset instead of `clang-linux` with a version `win` which is rarely what one wants.

It still doesn't work because then `clang-win` complains of missing `<compatibility>`, and I see no way of supplying that from the command line. Ideally, we should strive for `toolset=clang-win` working out of the box, but I don't know how feasible this is.

### DIFF
--- a/src/tools/clang.jam
+++ b/src/tools/clang.jam
@@ -17,7 +17,12 @@ feature.subfeature toolset clang : platform : : propagated link-incompatible ;
 
 rule init ( * : * )
 {
-    if [ os.name ] = MACOSX
+    if $(1) = win
+    {
+        toolset.using clang-win :
+          $(2) : $(3) : $(4) : $(5) : $(6) : $(7) : $(8) : $(9) ;
+    }
+    else if [ os.name ] = MACOSX
     {
         toolset.using clang-darwin : 
           $(1) : $(2) : $(3) : $(4) : $(5) : $(6) : $(7) : $(8) : $(9) ;


pdimov:
Any objections to this one?

eldiener:
Since Boost Build comes with a clang-win.jam shouldn't toolset=clang-win go directly to the init routine in that file ? Or does b2 only parse this as toolset=clang version=win and therefore use your PR logic above ?

pdimov:
That is what happens for me, yes. Without this patch, `toolset=clang-win` initializes `clang-linux` with a version `win`.

eldiener:
I am good with this PR.

pdimov:
Status quo is that `clang-linux` works (even when targeting llvm-win) and `clang-win` does not, so it's not yet appropriate to switch the default. -)

pdimov:
Well?

pdimov:
I think that dispatching when the version is `win`, `linux` or `darwin` is correct. This is the case when `toolset=clang-win` is explicitly given and misparsed.

If none of those is the case, that is, `toolset=clang` is given, we probably should switch to defaulting to `clang-win` on Windows, in due time.

pdimov:
The mingw clang works when using clang-linux, and there's no real need to force clang-win (which is msvc-derived) to also contain the functionality of clang-linux (which is gcc-derived).

These two toolsets are more properly called clang-msvc and clang-gcc because that's what they are; the platform itself doesn't really matter.

swatanabe:
AMDG

On 10/27/2018 11:55 AM, Rene Rivera wrote:
> Right.. Definitely trying to respond to the version as win, linux, or darwin is the way to go. But we need to test what happens for `clang-win-8` or `clang-win-7`, etc. As those should also correctly pick the platform and version aspects.
> 
> As for the default, we need some meaningful default based on the platform. So defaulting to the corresponding platform specific one based on OS seems reasonable.

Could we instead check
a) If a compiler path is passed, identify it as clang or clang-cl
b) If no compiler path is given, search for clang-cl and clang++
   in PATH (in that order for os.name = NT) and choose the
   appropriate toolset.

> But as before, the problem child is mingw. Which, IMO, in an ideal world `clang-win.jam` would also handle.
> 

That doesn't make any sense from an implementation POV.
If we're going to go down that route, we should just get
rid of the different clang-*.jam and have a single clang
toolset.

In Christ,
Steven Watanabe

eldiener:
On Windows clang targeting mingw-64/gcc and clang targeting vc++ are two entirely different compilers. Insisting that they both should be the same does not make them the same. Trying to corral both into the same .jam file seems wrong to me.

PR #331, opened 2018-08-19
Add visibility feature

Lastique:
The new visibility feature can be used to specify default symbol visibility
on compilers and platforms that support it. The default visibility is
hidden, unlike most compilers' defaults. This is intentional so that all
Boost libraries are compiled with hidden visibility by default. Other modes
are: protected and global (the latter is called "default" in gcc documentation).

I've only tested it on Linux, with gcc and clang.

The discussion that led to this PR is here:

http://boost.2283326.n4.nabble.com/all-Request-for-out-of-the-box-visibility-support-tt4704134.html

### DIFF
--- a/src/tools/features/visibility-feature.jam
+++ b/src/tools/features/visibility-feature.jam
@@ -1,0 +1,10 @@
+# Copyright 2018 Andrey Semashev
+# Distributed under the Boost Software License, Version 1.0.
+# (See accompanying file LICENSE_1_0.txt or copy at
+# http://www.boost.org/LICENSE_1_0.txt)
+
+import feature ;
+
+feature.feature visibility
+    : global protected hidden
+    : propagated optional ;
--- a/doc/src/reference.adoc
+++ b/doc/src/reference.adoc
@@ -522,6 +522,26 @@ in contexts where conditional properties are not allowed
 tell Boost.Build that the feature is actually used.
 ** Generators and targets that manipulate property-sets directly. Solution: set <relevant> manually.
 
+[[bbv2.builtin.features.visibility]]`visibility`::
+*Allowed values:* `global`, `protected`, `hidden`.
++
+This feature is used to specify the default symbol visibility in compiled
+binaries. Not all values are supported on all platforms and on some
+platforms (for example, Windows) symbol visibility is not supported at all.
++
+The supported values have the following meaning:
++
+* `global` - a.k.a. "default" in gcc documentation. Global symbols are considered
+  public, they are exported from shared libraries and can be redefined by another
+  shared library or executable. This is the default value.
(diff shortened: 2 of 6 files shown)


src/tools/features/visibility-feature.jam
```diff
@@ -0,0 +1,10 @@
+# Copyright 2018 Andrey Semashev
+# Distributed under the Boost Software License, Version 1.0.
+# (See accompanying file LICENSE_1_0.txt or copy at
+# http://www.boost.org/LICENSE_1_0.txt)
+
+import feature ;
+
+feature.feature visibility
+    : hidden protected global
```
grafikrobot: Setting the default to be different than the for compilers will affect everyone, not jus boost builds. Please change the default to match the compiler defaults. If boost wants a different kind of build it can set the feature in the appropriate jamfiles.

grafikrobot: Additionally, this needs to be an `optional` feature so that the compile option doesn't appear when not specified. I.e. which should be the default.

src/tools/features/visibility-feature.jam
```diff
@@ -0,0 +1,10 @@
+# Copyright 2018 Andrey Semashev
+# Distributed under the Boost Software License, Version 1.0.
+# (See accompanying file LICENSE_1_0.txt or copy at
+# http://www.boost.org/LICENSE_1_0.txt)
+
+import feature ;
+
+feature.feature visibility
+    : global protected hidden
+    : propagated ;
```
grafikrobot: Additionally, this needs to be an optional feature so that the compile option doesn't appear when not specified. I.e. which should be the default.

Lastique:
I've updated the feature and the docs according the comments, although it sort of defeats the purpose. We could modify library jamfiles before, the point was to make hidden visibility the default so that we don't have to do that. At least the property offers a more portable way to do that.

swatanabe:
AMDG

On 08/19/2018 11:47 AM, Andrey Semashev wrote:
> I've updated the feature and the docs according the comments, although it sort of defeats the purpose. We could modify library jamfiles before, the point was to make hidden visibility the default so that we don't have to do that. At least the property offers a more portable way to do that.
> 
> 

It can be set in the Boost Jamroot:

project /boost : ... : default-build <visibility>hidden ;

In Christ,
Steven Watanabe

Lastique:
Thanks, Steven, that sounds like a good idea. I'll create a PR for the superproject if this PR is accepted.

Lastique:
The related superproject PR is https://github.com/boostorg/boost/pull/190.

apolukhin:
Please merge to master. Used that feature in tests, now they fail on master (but do work perfectly on develop)

PR #332, opened 2018-08-20
boost/build: fix ch declaration in debugger.c

mksully22:
fix declaration of ch variable in debug_parent_wait() to be an int rather than a char. fgetc does return an int and if return variable is declared to be a char truncation occurs that results in later comparisons of ch to EOF to fail (0xff compare to 0xffffffff) on ppc64le.


swatanabe:
AMDG

Thanks!  I noticed the failures in the regression
matrix before, but I couldn't track down the
problem.

In Christ,
Steven Watanabe

PR #334, opened 2018-08-20
add simple clang_tidy toolset

arvidn:
This makes it possible to run clang-tidy instead of building. instead of creating object files and binaries, it simply just creates empty files, to indicate it was successfully run.

I've been using this in my project to run clang-tidy on travis, [here](https://github.com/arvidn/libtorrent/blob/master/.travis.yml#L123). (note that compilation errors are not reported as failures by clang-tidy, which happens to make that work because clang-tidy on travis has some issues with the standard library I believe).

The reason this toolset is not called `clang-tidy` is that it would clash with the clang toolset, since `-` is used to separate toolset name from version.

It does not handle `run` rules very gracefully, as the binaries it outputs are just empty files, trying to run them will fail. But it works as long as you specify `testing.execute=off`.

I think this is a much better approach than https://github.com/boostorg/build/pull/133, and is meant to replace it.

Thoughts?

### DIFF
--- a/src/tools/clang_tidy.jam
+++ b/src/tools/clang_tidy.jam
@@ -1,0 +1,81 @@
+#  Copyright (c) 2018 Arvid Norberg (arvid@libtorrent.org)
+#
+#  Use, modification and distribution is subject to the Boost Software
+#  License Version 1.0. (See accompanying file LICENSE_1_0.txt or
+#  http://www.boost.org/LICENSE_1_0.txt)
+
+import common ;
+import toolset ;
+import feature ;
+
+feature.extend toolset : clang_tidy ;
+
+generators.register-c-compiler clang_tidy.compile.c++ : CPP : OBJ : <toolset>clang_tidy ;
+generators.register-c-compiler clang_tidy.compile.c : C : OBJ : <toolset>clang_tidy ;
+generators.register-archiver clang_tidy.archive : OBJ : STATIC_LIB : <toolset>clang_tidy ;
+generators.register-linker clang_tidy.link : OBJ SEARCHED_LIB STATIC_LIB : EXE : <toolset>clang_tidy ;
+generators.register-linker clang_tidy.link.dll : OBJ SEARCHED_LIB STATIC_LIB : SHARED_LIB : <toolset>clang_tidy ;
+
+rule init ( version ? :  command * : options * ) {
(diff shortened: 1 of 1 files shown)


src/tools/clang_tidy.jam
```diff
@@ -0,0 +1,81 @@
...
+
+toolset.flags clang_tidy.compile OPTIONS <rtti>off : -fno-rtti ;
+
+###############################################################################
+# C and C++ compilation
+
+TOUCH = [ common.file-creation-command ] ;
+
+actions compile.c++ {
+  "$(CONFIG_COMMAND)" -header-filter=*. -warnings-as-errors=* "$(>)" -- -x c++ $(OPTIONS) -D$(DEFINES) -I"$(INCLUDES)" \
```
frenchtoast747: Does clang-tidy allow for response files as input?

If so, it would be nice to generate that here in the event that `DEFINES` or `INCLUDES` becomes large (to prevent blowing through max command line length):
e.g. `@@($(>:D).rsp:E=$(OPTIONS) -D$(DEFINES) -I"$(INCLUDES)")`

See [msvc.jam](https://github.com/boostorg/build/blob/develop/src/tools/msvc.jam#L605) for an example of the syntax and note that you may have to join each of the options with a newline depending on how clang's response files work.

Same for `compile.c` below.

src/tools/clang_tidy.jam
```diff
@@ -0,0 +1,81 @@
...
+feature.extend toolset : clang_tidy ;
+
+generators.register-c-compiler clang_tidy.compile.c++ : CPP : OBJ : <toolset>clang_tidy ;
+generators.register-c-compiler clang_tidy.compile.c : C : OBJ : <toolset>clang_tidy ;
+generators.register-archiver clang_tidy.archive : OBJ : STATIC_LIB : <toolset>clang_tidy ;
+generators.register-linker clang_tidy.link : OBJ SEARCHED_LIB STATIC_LIB : EXE : <toolset>clang_tidy ;
+generators.register-linker clang_tidy.link.dll : OBJ SEARCHED_LIB STATIC_LIB : SHARED_LIB : <toolset>clang_tidy ;
+
+rule init ( version ? :  command * : options * ) {
+  command = [ common.get-invocation-command clang_tidy : clang-tidy
```
frenchtoast747: It would be nice if this could provide some common installation paths (if there are any).

PR #336, opened 2018-08-28
Add VxWorks process extension.

kuhlenough:
We use "vxe" extension for VxWorks processes by convention, no tool requires it, its just an easy way to identify the cross built executable.


kuhlenough:
Where are these sanity tests stored, perhaps I can contribute some updates to make them a little more realistic?   
BTW.  If I had a kernel  on QEMU could Travis or AppVeyor, start it so we could do some run testing?

PR #341, opened 2018-09-23
build: misc.typos

luzpaz:
Found via `codespell -q 3 --skip="./src/engine/boehm_gc" -L te,iff`


luzpaz:
Apologies, should be reverted in 18880d8

PR #350, opened 2018-10-10
Use correct sizeof in malloc call

jwakely:
This is allocating space for `nel` objects of type `ITEM*` so it should use `sizeof(ITEM*)` not `sizeof(ITEM**)`.

In practice the values are the same, but using the correct type is better anyway, and now matches the same calculation in the `memset` call in the following statement.


mclow:
This looks good to me.

PR #360, opened 2018-10-27
Fix clang-win name mangling

pdimov:
It previously had no tag, so f.ex. Clang 6.0.1 just added "-60" to the mangled name instead of "-clangw6" as now.

### DIFF
--- a/src/tools/common.jam
+++ b/src/tools/common.jam
@@ -918,6 +918,7 @@ local rule toolset-tag ( name : type ? : property-set )
             {
                case darwin : tag += clang-darwin ;
                case linux  : tag += clang ;
+               case win    : tag += clangw ;
             }
         }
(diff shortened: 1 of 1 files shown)


eldiener:
I realized I am a little worried about this as clang on Windows targeting mingw-64/gcc would have the same tag as clang on Windows targeting vc++. I do not know how the tag is used but those are two different compilers in my opinion.

pdimov:
It doesn't; the mingw/Cygwin clang is clang-linux, and has tag `clang`, and clang-cl is clang-win, and has tag `clangw`.

The tag is encoded into the library name, f.ex. `boost_atomic-clangw6-mt-d-x64-1_69.dll`.

eldiener:
Thanks for the explanation of how the tag is used. The code looks fine.

Issue #330, opened 2018-08-13
The clang-win toolset doesn't work


pdimov:
Trying to use the `clang-win` toolset fails with

```
C:/boost-git/develop/tools/build/src/tools\msvc.jam:1087: in set-setup-command
*** argument error
* rule virtual-target.from-file ( file : file-loc : project )
* called with: (  : /C:/boost-git/develop/libs/assert : object(project-target)@129 )
* missing argument file
C:/boost-git/develop/tools/build/src/build\virtual-target.jam:970:see definition of rule 'virtual-target.from-file' being called
C:/boost-git/develop/tools/build/src/tools\msvc.jam:652: in clang-win.compile.c++
C:/boost-git/develop/tools/build/src/kernel\modules.jam:107: in modules.call-in
C:/boost-git/develop/tools/build/src/util\indirect.jam:105: in indirect.call
C:/boost-git/develop/tools/build/src/build\virtual-target.jam:902: in execute
C:/boost-git/develop/tools/build/src/build\virtual-target.jam:821: in class@action.actualize
C:/boost-git/develop/tools/build/src/build\virtual-target.jam:332: in actualize-action
C:/boost-git/develop/tools/build/src/build\virtual-target.jam:518: in actualize-no-scanner
C:/boost-git/develop/tools/build/src/build\virtual-target.jam:142: in class@virtual-target.actualize
C:/boost-git/develop/tools/build/src/build\configure.jam:258: in try-find-build
C:/boost-git/develop/tools/build/src/build\configure.jam:396: in find-builds-raw
C:/boost-git/develop/tools/build/src/build\configure.jam:455: in configure.find-builds
C:/boost-git/develop\boostcpp.jam:734: in boostcpp.deduce-address-model
C:/boost-git/develop/tools/build/src/kernel\modules.jam:107: in modules.call-in
C:/boost-git/develop/tools/build/src/util\indirect.jam:105: in indirect.call
C:/boost-git/develop/tools/build/src/build\property.jam:144: in property.evaluate-conditionals-in-context
C:/boost-git/develop/tools/build/src/build\targets.jam:1087: in evaluate-requirements
C:/boost-git/develop/tools/build/src/build\targets.jam:1121: in common-properties2
C:/boost-git/develop/tools/build/src/build\targets.jam:1017: in targets.common-properties
C:/boost-git/develop/tools/build/src/build\targets.jam:1313: in class@basic-target.generate
[...]

PR #361, opened 2018-10-27
Fix clang-win.jam so that it works

pdimov:


pdimov:
Fixes #330.

pdimov:
With this in place, clang-cl should now be usable on Appveyor (which has it installed and in PATH), by just using `toolset=clang-win`.

eldiener:
The new clang-win.jam worked well. Is there a way to choose a VC++ version as the backend headers and libraries, or is the latest VC++ version found always chosen ? The old clang-win had a <compatibility> property but the new one does not have anything like that which I can see.

pdimov:
I'm not sure how `clang-cl` picks the version to use. Will need to check the sources. I think that if you invoke it from a developer command prompt, it uses whatever is set by it, but if not, it scans and picks the most recent MSVC toolset (and Windows SDK - these two are separate.)

PR #369, opened 2018-11-17
add clang to the list of supported toolsets for bootstrap.sh

jeking3:
clang was not in the list of supported toolsets - it was also missing from the Guess_Toolset script so I slotted it in under gcc.  This means when guessing we will prefer gcc, then clang.  It is better than missing clang all together and being unable to guess.

clang is detected a couple lines earlier for FreeBSD, which means if you are on FreeBSD and you have clang it is used, otherwise gcc.  I'm not sure this is really necessary here, but I didn't change it.


pdimov:
clang is the system compiler on FreeBSD, that's probably why.

PR #370, opened 2018-11-17
gcc compiler warnings identified unhandled error cases

jeking3:
error on fork had no output and did not end correctly


src/engine/debugger.c
```diff
@@ -1233,6 +1244,8 @@ static void debug_start_child( int argc, const char * * argv )
         close( pipe1[ 1 ] );
         close( pipe2[ 0 ] );
         close( pipe1[ 1 ] );
```
jeking3: Didn't see this before, but there was a typo on this line that should be fixed.  It was already there, not part of this PR.

PR #375, opened 2018-12-02
 Unify mips architectures

FlyGoat:


FlyGoat:
These four PRs should be accepted together:
- https://github.com/boostorg/build/pull/375
- https://github.com/boostorg/config/pull/254
- https://github.com/boostorg/context/pull/93
- https://github.com/boostorg/log/pull/66

FlyGoat:
> 1. Adding the general "mips" architecture without removing the rest. We would need to indicate they non-general as deprecated somehow.

Thanks for your suggestions, where should be the best place to add such a warning?

It would be better if these PRs can be accepted at once so that build the whole boost on mips won't be broken.

Lastique:
@FlyGoat Is this no longer relevant? Should the other PRs be closed as well?

FlyGoat:
@Lastique Sorry my fault, I'll rebase and reopen it. Yes, it's still a problem.

FlyGoat:
@wzssyqa handed will sucess this PR.

PR #392, opened 2019-02-16
boostbook: Fixed DTDXML path problems

Kojoley:


Kojoley:
I would appreciate if you point me how I could make this:

```
boostbook reference : reference.dtdxml : <location>. ;
explicit reference ;

boostbook boostbook : boostbook.xml : <dependency>reference ;
```
without `<location>.` (to not pollute working tree).

Something like:
```
boostbook reference : reference.dtdxml : : : <xsl:path>path/to/this/target/location ;
```

PR #393, opened 2019-02-18
Fix scanner regex pattern

jonathanpoelen:
Before, commented includes are added as dependencies

```cpp
#include "a.h" // recompiled if a.h is updated: ok
// #include "b.h" // recompiled if b.h is updated: bad
int main() {}
```

The new pattern checks that `#` is preceded by 0 or more blank characters


jonathanpoelen:
I am moderately convinced by this approach. In fact, why use a regex rather than a compiler mechanisms that lists includes?

Issue #382, opened 2019-01-06
"bootstrap.bat borland" fails to find label "Test_Path"


tanzislam:
I have both VS2017 and Embarcadero C++Builder 10.3 (i.e. the `borland` toolchain) installed in my Windows 7 SP1 (64-bit) laptop. I freshly cloned the Boost super-project with all submodules (`git submodule update --init --recursive`), then tried building Boost.Build specifically with the `borland` toolchain in a plain `cmd.exe` window:

```
>bootstrap.bat borland
```

This failed with:

```
Building Boost.Build engine
The system cannot find the batch label specified - Test_Path

Failed to build Boost.Build engine.
Please consult bootstrap.log for further diagnostics.
```

tanzislam:
I believe the reason is that the `Test_Path` label exists inside `src/engine/guess_toolset.bat`, but that script is skipped in the call path (`bootstrap.bat` -> `src/engine/build.bat` -> `src/engine/config_toolset.bat`) if a toolset argument is provided. The error occurs at the `:Skip_VC141` logic in `config_toolset.bat` when it calls `Test_Path`.

PR #398, opened 2019-02-23
Fix build with Embarcadero C++ Builder ("borland" toolset)

tanzislam:
Fix several problems with building with the `borland` toolset on a Embarcadero C++ Builder 10.3.1 ([Community Edition](https://www.embarcadero.com/products/cbuilder/starter)) installation:

 - `bootstrap.bat` -> `src/engine/build.bat` bypasses loading `src/engine/guess_toolset.bat` if an override toolset is given. This later leads to the call to `Test_Path` failing. (Issue #382)
 - C++Builder toolchain path is intolerant of spaces, e.g. `C:\Program Files\Embarcadero\…`
 - Old Borland (`bcc32.exe`) preprocessor doesn't accept `#if SYMBOL_NAME`, but insists on `#if defined(SYMBOL_NAME)` or `#ifdef SYMBOL_NAME`. Work around this by providing `-Nd` to auto-define `SYMBOL_NAME` to `1`.

With these three fixes, running `bootstrap.bat borland` in the super-project directory should successfully produce a working `b2.exe`.

NOTE: To actually build some Boost libraries with the `borland` toolset, we also need a `user-config.jam` file to force the use of `bcc32c.exe` for modern C++ support:

```
using borland : 7.40 : bcc32c.exe -DBOOST_STRICT_CONFIG -DBOOST_ASSERT_CONFIG ;
```

(I've [managed to build](https://github.com/tanzislam/cryptopals/wiki/Importing-into-Embarcadero-C┼┼Builder-Starter-10.2#linking-in-a-boost-library) Boost.System, Boost.DateTime and Boost.Regex in this manner.)

### DIFF
--- a/src/engine/build.bat
+++ b/src/engine/build.bat
@@ -100,8 +100,8 @@ if errorlevel 1 (
 
 call guess_toolset.bat
 if errorlevel 1 (
-    call :Error_Print "Could not find a suitable toolset."
-    goto :eof)
+    call :Error_Print "Could not find a suitable toolset.")
+goto :eof
 
 
 :Guess_Yacc
--- a/src/engine/build.jam
+++ b/src/engine/build.jam
@@ -176,7 +176,7 @@ toolset acc cc : "-o " : -D
     -I$(--python-include) -I$(--extra-include)
     : -L$(--python-lib[1]) -l$(--python-lib[2]) ;
 ## Borland C++ 5.5.x
-toolset borland bcc32 : -e -n : /D
+toolset borland bcc32 : -e -n : "-Nd /D" " /D" ""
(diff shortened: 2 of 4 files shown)


tanzislam:
Hmm … not sure what went wrong with the AppVeyor build. I could not reproduce the issue on my side with the following in a plain `cmd.exe` window:

```
git clone https://github.com/tanzislam/build.git
cd build
git checkout origin/fix-borland-build
set PATH=C:\Windows\system32;C:\Windows;C:\Windows\system32\Wbem
cd src\engine
build.bat
```

(I have VS2017 Community v15.9.7 and Git for Windows v2.19.2.windows.1 installed in my Windows 7 x64 SP1 laptop.)

Issue #376, opened 2018-12-13
b2 layout command line parameter is not docummented


SSE4:
I wasn't able to find documentation for b2 layout command line parameter.
it seems like it's not even mentioned anywhere (https://www.google.com/search?q=site%3Ahttps%3A%2F%2Fboostorg.github.io%2Fbuild%2F+layout&ie=utf-8&oe=utf-8)

the following page https://www.boost.org/doc/libs/1_69_0/more/getting_started/windows.html seems to be having lots of broken links (404), for instance:

- For a complete description of these and other invocation options, please see the [Boost.Build documentation](https://boostorg.github.io/build/doc/html/bbv2/overview/invocation.html).
- an up-to-date list is always available in the [Boost.Build documentation](https://boostorg.github.io/build/doc/html/bbv2/reference/tools.html)
- If you previously chose a toolset for the purposes of [building b2](https://www.boost.org/doc/libs/1_69_0/doc/html/bbv2.html#bbv2.installation), you should assume it won't work and instead choose newly from the table below.

this is very inconvenient. the only option to get some documentation, is to use `b2 --help`:
```
 --layout=<layout>       Determine whether to choose library names and header
                          locations such that multiple versions of Boost or
                          multiple compilers can be used on the same system.

                              -- versioned -- Names of boost binaries include
                              the Boost version number, name and version of
                              the compiler and encoded build properties. Boost
                              headers are installed in a subdirectory of
                              <HDRDIR> whose name contains the Boost version
                              number.

                              -- tagged -- Names of boost binaries include the
                              encoded build properties such as variant and
                              threading, but do not including compiler name
                              and version, or Boost version. This option is
                              useful if you build several variants of Boost,
                              using the same compiler.

                              -- system -- Binaries names do not include the
                              Boost version number or the name and version
                              number of the compiler. Boost headers are
                              installed directly into <HDRDIR>. This option is
                              intended for system integrators building
                              distribution packages.

```
however, this obviously requires to have b2 executable (which might be missing, e.g. if you're on Mobile, or using boost from package manager, or whatever).

also, this documentation from `b2 --help` is very brief, as it doesn't fully describe on how final libraries/binaries will be named, there is no naming template, no examples, no specification on how "encoded build properties" are actually encoded, etc.

/cc @apolukhin @grafikrobot

grafikrobot:
[...]

PR #400, opened 2019-02-28
Include --layout option in documentation

Adam-Gleave:
Simple change to include the `--layout` option in the documentation (under Overview->Options). I was searching for a while to figure out how to alter the naming of the built libraries because the docs did not outline anything. Hopefully this should help someone in a similar position.

### DIFF
--- a/doc/src/overview.adoc
+++ b/doc/src/overview.adoc
@@ -587,6 +587,18 @@ Boost.Build recognizes the following command line options.
   Paths”].
 `--version`::
   Prints information on the Boost.Build and Boost.Jam versions.
+`--layout`::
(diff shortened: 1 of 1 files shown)


swatanabe:
AMDG

On 2/28/19 4:05 AM, Adam Gleave wrote:
> Simple change to include the `--layout` option in the documentation (under Overview->Options). I was searching for a while to figure out how to alter the naming of the built libraries because the docs did not outline anything.

There is no such option in b2 itself.  See also #376.

In Christ,
Steven Watanabe

Adam-Gleave:
No problem, that makes sense. I'll take a look at the Boost documentation itself and see if it needs this.
Thanks.

PR #403, opened 2019-03-05
If the vswhere lookup fails, keep searching for others

teeks99:
Exiting with a non-zero errorlevel caused it to exit from guess_toolset.cmd, we want it to keep looking for other versions.

How did bootstrap keep working on machines without VS2017/2019 installed?

### DIFF
--- a/src/engine/guess_toolset.bat
+++ b/src/engine/guess_toolset.bat
@@ -32,21 +32,20 @@ goto :eof
 REM Check the variable first. This can be set manually by the user (by running the tools command prompt).
 call :Clear_Error
 call vswhere_usability_wrapper.cmd
-if NOT "_%VS150COMNTOOLS%_" == "__" (
-    set "BOOST_JAM_TOOLSET=vc141"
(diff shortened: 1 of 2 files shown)


teeks99:
Let me test this a bit more before merging, I just wanted to get it out there for the sake of conversation in #401

teeks99:
This fixes what happens when vswhere fails. 

Tested with:
*  VS2019 only machine, cmd.exe, vswhere
*  VS2019 only machine, 2019 command prompt, vswhere, VS160COMNTOOLS set
*  VS2019 only machine, 2019 command prompt, no-vswhere, VS160COMNTOOLS set (Fixes this case!)
*  VS2015 only machine, cmd.exe VS140COMNTOOLS set
*  Machine with vs2010 through 2019, cmd.exe (VS100, VS110, VS120, VS140 COMNTOOLS set)

In all cases, bootstrap.bat completed successfully.

swatanabe:
AMDG

Aha!  That makes more sense than your first patch.
Looks good to me.

In Christ,
Steven Watanabe

PR #406, opened 2019-03-10
Better doc

jehelset:
* doc/src/reference.adoc
  ( bbv2.reference.init ): Mention Unix specific BOOST_BUILD_PATH (https://github.com/boostorg/build/blob/develop/src/engine/Jambase#L48-L55)


jehelset:
i think perhaps the travis ci fail was a spurious error.

PR #409, opened 2019-03-10
Support for common sanitizers

djarek:
Add support for a common set of sanitizers supported by both gcc and clang:
- thread
- address
- undefined

### DIFF
--- a/src/tools/gcc.jam
+++ b/src/tools/gcc.jam
@@ -731,9 +731,9 @@ toolset.flags gcc.compile.c++ OPTIONS <exception-handling>off : -fno-exceptions 
 toolset.flags gcc.compile.c++ OPTIONS <address-sanitizer>on : -fsanitize=address ;
 toolset.flags gcc.compile.c++ OPTIONS <undefined-sanitizer>on : -fsanitize=undefined ;
 toolset.flags gcc.compile.c++ OPTIONS <thread-sanitizer>on : -fsanitize=thread ;
-toolset.flags gcc.compile.c++ OPTIONS <address-sanitizer>on-no-recover : -fsanitize=address -fno-sanitize-recover=address ;
-toolset.flags gcc.compile.c++ OPTIONS <undefined-sanitizer>on-no-recover : -fsanitize=undefined -fno-sanitize-recover=undefined ;
-toolset.flags gcc.compile.c++ OPTIONS <thread-sanitizer>on-no-recover : -fsanitize=thread -fno-sanitize-recover=thread ;
+toolset.flags gcc.compile.c++ OPTIONS <address-sanitizer>norecover : -fsanitize=address -fno-sanitize-recover=address ;
+toolset.flags gcc.compile.c++ OPTIONS <undefined-sanitizer>norecover : -fsanitize=undefined -fno-sanitize-recover=undefined ;
+toolset.flags gcc.compile.c++ OPTIONS <thread-sanitizer>norecover : -fsanitize=thread -fno-sanitize-recover=thread ;
 
 # configure Dinkum STL to match compiler options
 toolset.flags gcc.compile.c++ DEFINES <rtti>off/<target-os>vxworks : _NO_RTTI ;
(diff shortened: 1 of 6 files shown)


mclow:
I think having an example in the docs will make this much better.

mclow:
Thanks! Looks good to me.

swatanabe:
AMDG

- Please don't use a '-' in a feature value.  `-`
  is used to separate subfeatures.
- Since the feature is optional, you don't need
  an `off` value.  The absence of the feature
  means `off`.

In Christ,
Steven Watanabe

Kojoley:
The thing is that you can supply multiple sanitizes like `-fsanitize=address,undefined`, but the proposed method will not work for multiple sanitizes because `-fsanitize=address -fsanitize=undefined` is not allowed.

djarek:
@Kojoley you sure about that? Both clang and gcc seem to be fine with `-fsanitize=address -fsanitize=undefined`.

djarek:
@swatanabe 
Anything else to fix?

Kojoley:
> @Kojoley you sure about that? Both clang and gcc seem to be fine with -fsanitize=address -fsanitize=undefined.

It seems I was wrong. Probably I tried to use other set of flags or it was fixed.

src/tools/gcc.jam
```diff
@@ -868,6 +874,12 @@ toolset.flags gcc.link OPTIONS <local-visibility>hidden : -fvisibility=hidden -f
 toolset.flags gcc.link OPTIONS <local-visibility>protected : -fvisibility=protected ;
 toolset.flags gcc.link OPTIONS <local-visibility>protected/<target-os>darwin : ;
 toolset.flags gcc.link OPTIONS <local-visibility>global : -fvisibility=default ;
+toolset.flags gcc.link OPTIONS <address-sanitizer>on : -fsanitize=address ;
+toolset.flags gcc.link OPTIONS <undefined-sanitizer>on : -fsanitize=undefined ;
+toolset.flags gcc.link OPTIONS <thread-sanitizer>on : -fsanitize=thread ;
+toolset.flags gcc.link OPTIONS <address-sanitizer>norecover : -fsanitize=address -fno-sanitize-recover=address ;
+toolset.flags gcc.link OPTIONS <undefined-sanitizer>norecover : -fsanitize=undefined -fno-sanitize-recover=undefined ;
+toolset.flags gcc.link OPTIONS <thread-sanitizer>norecover : -fsanitize=thread -fno-sanitize-recover=thread ;
```
Kojoley: IIRC `-fno-sanitize-recover` is not a liker flag, but compile flag.

djarek:
@grafikrobot I think there's nothing left to change.

src/tools/gcc.jam
```diff
@@ -728,6 +728,12 @@ toolset.flags gcc.compile OPTIONS <local-visibility>global : -fvisibility=defaul
 
 toolset.flags gcc.compile.c++ OPTIONS <rtti>off : -fno-rtti ;
 toolset.flags gcc.compile.c++ OPTIONS <exception-handling>off : -fno-exceptions ;
+toolset.flags gcc.compile.c++ OPTIONS <address-sanitizer>on : -fsanitize=address ;
+toolset.flags gcc.compile.c++ OPTIONS <undefined-sanitizer>on : -fsanitize=undefined ;
+toolset.flags gcc.compile.c++ OPTIONS <thread-sanitizer>on : -fsanitize=thread ;
+toolset.flags gcc.compile.c++ OPTIONS <address-sanitizer>norecover : -fsanitize=address -fno-sanitize-recover=address ;
+toolset.flags gcc.compile.c++ OPTIONS <undefined-sanitizer>norecover : -fsanitize=undefined -fno-sanitize-recover=undefined ;
```
jeking3: The description says "for gcc and clang", so... per clang documentation on this issue:

https://clang.llvm.org/docs/AddressSanitizer.html

To get a reasonable performance add -O1 or higher. To get nicer stack traces in error messages add -fno-omit-frame-pointer. To get perfect stack traces you may need to disable inlining (just use -O1) and tail call elimination (-fno-optimize-sibling-calls).

I don't ususally add -O1, but I do add -g and -fno-omit-frame-pointer, and never tried -fno-optimize-sibling-calls.

src/tools/gcc.jam
```diff
@@ -728,6 +728,12 @@ toolset.flags gcc.compile OPTIONS <local-visibility>global : -fvisibility=defaul
 
 toolset.flags gcc.compile.c++ OPTIONS <rtti>off : -fno-rtti ;
 toolset.flags gcc.compile.c++ OPTIONS <exception-handling>off : -fno-exceptions ;
+toolset.flags gcc.compile.c++ OPTIONS <address-sanitizer>on : -fsanitize=address ;
+toolset.flags gcc.compile.c++ OPTIONS <undefined-sanitizer>on : -fsanitize=undefined ;
+toolset.flags gcc.compile.c++ OPTIONS <thread-sanitizer>on : -fsanitize=thread ;
```
jeking3: Missed support for -fsanitize=integer

src/tools/gcc.jam
jeking3: It also helps to define the environment variable UBSAN_OPTIONS=print_stacktrace=1 when running with the sanitizers, at least with the undefined one; I'm not sure this applies to other ones.

PR #415, opened 2019-03-24
Fix failure when $CC contained spaces

pdimov:


pdimov:
```
build.jam:187: Unescaped special character in argument $(CC:J= )
```

That's overly pedantic; the `:` is inside `()`, so it shouldn't be interpreted as a separator.

swatanabe:
AMDG

On 3/24/19 7:39 PM, Peter Dimov wrote:
> ```
> build.jam:187: Unescaped special character in argument $(CC:J= )
> ```
> 
> That's overly pedantic; the `:` is inside `()`, so it shouldn't be interpreted as a separator.
> 

It's not the ':'.  The reason that this happens is
probably because of the space.  This would previously
have been ill-formed because it was parsed as "$(CC:J=" ")"
I admit the message is a bit misleading here and is
probably unnecessary.

In Christ,
Steven Watanabe

khourig:
So close.

PR #423, opened 2019-04-03
help: Print help for local modules as well.

tee3:
This pull request adds support for help for modules in the current directory.  I added this because using `--help-internal` did not list the module and `--help module` with did not show the help.  I don't know if this is the best approach to adding this functionality, but adding new options would have required changing a lot of code in `help.jam` and `doc.jam`.  It was not simple to just add something there without duplicating some complicated code.

Both the `--help-internal` and `--help module` now search the current directory for modules.

I originally thought that adding the current directory to `BOOST_BUILD_PATH` would solve this problem, but it did not seem to.

I tried to poke the current directory into the `BOOST_BUILD_PATH` in each of `user-config.jam`, `project-config.jam`, and`site-config.jam`.  This works for importing the local module, but not when using `--help`.

```
{
  import modules ;

  local x = [ modules.peek : BOOST_BUILD_PATH ] ;
  modules.poke : BOOST_BUILD_PATH : . $(x) ;
}
```

I'm was sure this is the best way to approach this, so I have a few questions.

1. Is there a different way to do this?  Am I totally off base?

2. Should modifying `BOOST_BUILD_PATH` locally work?  If so, should I be doing that another way?

3. Should we add new options such as `--help-local` and`--help-internal-local` instead of modifying the search path of `--help` and `--help-internal`?

4. Should we rework the code of `help.jam` and `doc.jam` to be a bit more flexible?

### DIFF
--- a/src/options/help.jam
+++ b/src/options/help.jam
@@ -48,7 +48,7 @@ rule process (
     {
         case --help-internal :
         local path-to-modules = [ modules.peek : BOOST_BUILD_PATH ] ;
-        path-to-modules ?= . ;
+        path-to-modules += . ;
         local possible-modules = [ GLOB $(path-to-modules) : *\\.jam ] ;
         local not-modules = [ GLOB $(path-to-modules) : *$(.not-modules)\\.jam ] ;
         local modules-to-list =
@@ -177,7 +177,7 @@ local rule split-symbol (
     )
 {
     local path-to-modules = [ modules.peek : BOOST_BUILD_PATH ] ;
-    path-to-modules ?= . ;
+    path-to-modules += . ;
     local module-name = $(symbol) ;
     local symbol-name = ;
     local result = ;


swatanabe:
AMDG

On 4/3/19 9:55 AM, tee3 wrote:
> This pull request adds support for help for modules in the current directory.  <snip>
> I originally thought that adding the current directory to `BOOST_BUILD_PATH` would solve this problem, but it did not seem to.
> 

It should work if BOOST_BUILD_PATH is set in the environment.

> I tried to poke the current directory into the `BOOST_BUILD_PATH` in each of `user-config.jam`, `project-config.jam`, and`site-config.jam`.  This works for importing the local module, but not when using `--help`.
> 

That's because --help bypasses the normal
build process including loading config files.

In Christ,
Steven Watanabe

tee3:
@swatanabe In your opinion, should I close this pull request?

> It should work if BOOST_BUILD_PATH is set in the environment.

Thanks.  I misunderstood how that environment variable worked and I thought setting it would override the default `BOOST_BUILD_PATH`.  It is not too bad to require that it is set it this way, though it's one more thing to describe to users.

> That's because --help bypasses the normal build process including loading config files.

I'm assuming this is in part to avoid errors that occur if it is run where this is no Boost.Build system.  Are there other reasons for this?  Is there a way to do this without bypassing the normal process?

swatanabe:
AMDG

On 4/3/19 11:09 AM, tee3 wrote:
> @swatanabe In your opinion, should I close this pull request?
> 

No.  I think it's probably a good idea for
--help to use the same path that's used
for finding modules.  Jamfiles automatically
get their own directory added to BOOST_BUILD_PATH
when they import things.  Adding the current
working directory isn't prefect, but it is an
improvement.

>> It should work if BOOST_BUILD_PATH is set in the environment.
> 
> Thanks.  I misunderstood how that environment variable worked and I thought setting it would override the default `BOOST_BUILD_PATH`.  It is not too bad to require that it is set it this way, though it's one more thing to describe to users.
> 
>> That's because --help bypasses the normal build process including loading config files.
> 
> I'm assuming this is in part to avoid errors that occur if it is run where this is no Boost.Build system.  Are there other reasons for this?

The main reason is that it's not necessary.
We don't want to build anything.  The config
files normally don't do anything that would
affect the behavior of --help.

>  Is there a way to do this without bypassing the normal process?
> 

--help gets processed very early via a generic
option processing hook, so I don't think this
would be easy.

In Christ,
Steven Watanabe

PR #428, opened 2019-04-15
Fix build on KUbuntu 19.04

djarek:
Bootstrap fails on 19.04 because of references to non-existent `*.c` files, even though building `b2` succeeds.


djarek:
The file got removed, it's not working without it anyway.

PR #436, opened 2019-05-09
Clarify use of 'using' in Jamfiles vs configuration files

mloskot:


grisumbras:
While it is usable, wouldn't it prevent other people to configure used tools in some other way?

grisumbras:
It would only work that way, if the tool's module was written to handle it that way, correct?

grisumbras:
I just checked with `using gcc ;` and b2 errors out complaining about duplicate initialization.

swatanabe:
AMDG

On 5/9/19 6:17 AM, Dmitry wrote:
> It would only work that way, if the tool's module was written to handle it that way, correct? 
> 

Yep.  Some tools do handle it, just not compilers.  Incidentally,
the main reason that the compilers don't handle default initialization
is that they don't need to.  Boost.Build has hard-coded default compiler
initialization.  I would actually prefer it if we required Jamfiles to
have:

using cxx ;

or some such.

In Christ,
Steven Watanabe

grisumbras:
I checked the toolset module authoring guidelines and, IMO, they suggest that

```
using foo : 1 : ... ;
using foo : 2 : ... ;
using foo ;
```
should *not* work. 
So, maybe the guidelines should be changed a bit?

mloskot:
I'm not quite capable to provide any relevant input to the wider discussion that has rolled out.

> The using statement is actually perfectly usable in non-configuration source

Meanwhile, what about updating this PR and flip the note to read

```
TIP: You should prefer to use the  `using` rule in project's `Jamfile` file.
```

> Did you run into an instance where doing that didn't work?

No, I didn't. It was suggested to me in [this thread on Slack](https://cpplang.slack.com/archives/C27KZLB0X/p1557350198230000). It quite made sense to me, so I assumed it is correct and decided to add relevant update to the docs.

swatanabe:
AMDG

On 5/9/19 7:47 AM, Mateusz Łoskot wrote:
> I'm not quite capable to provide any relevant input to the wider discussion that has rolled out.
> 
>> The using statement is actually perfectly usable in non-configuration source
> 
> Meanwhile, what about updating this PR and flip the note to read
> 
> ```
> TIP: You should prefer to use the  `using` rule in project's `Jamfile` file.
> ```
> 

That's also wrong.  The current state is:
- Put `using tool ;` with no other argument in a Jamfile that
  needs `tool`, provided that the tool supports this usage.
- In all other cases, `using` should be in a configuration file.

The general principles behind this are:
a) Jamfiles should be portable.  Config files are system specific.
b) If you have a standard configuration (i.e. all tools are
   installed by your platform's package manager), config files
   should be unnecessary.

Note that this is only advisory.  Boost.Build doesn't enforce it.

>> Did you run into an instance where doing that didn't work?
> 
> No, I didn't. It was suggested to me in [this thread on Slack](https://cpplang.slack.com/archives/C27KZLB0X/p1557350198230000). It quite made sense to me, so I assumed it is correct and decided to add relevant update to the docs.
> 

In Christ,
Steven Watanabe

mloskot:
> The current state is:
> - Put `using tool ;` with no other argument in a Jamfile that needs `tool`, provided that the tool supports this usage.
> 
> - In all other cases, `using` should be in a configuration file. 

It makes perfect sense. Why not to add these notes to the docs? Even a single `TIP` would make things clearer, IMO.

mloskot:
@grafikrobot  I do care, but I'm attending a hackathon in Minneapolis, so may not be able to do it before I'm back home  next week.

mloskot:
@grafikrobot I've modified the note according to the revelations discussed in comments. Is it any better, anything to correct?

PR #449, opened 2019-06-08
simplified coverage options and added docs

HDembinski:
Thank you for working on the coverage feature. I discovered the previous work by accident. This patch simplifies the options to just "on" and "off". `-fprofile-arcs` could be used alone, but there is no use case known to me for instrumenting the code but then not write the results somehwere. `-ftest-coverage` doesn't work without `-fprofile-arcs`. I argue the options should be kept simple, unless more fine-grained control is actually requested, and then more options can be added on top of "on" and "off".

This patch also adds documentation.

### DIFF
--- a/src/tools/features/coverage-feature.jam
+++ b/src/tools/features/coverage-feature.jam
@@ -1,12 +1,13 @@
 # Copyright 2019 Rene Rivera
+# Copyright 2019 Hans Dembinski
 # Distributed under the Boost Software License, Version 1.0.
 # (See accompanying file LICENSE_1_0.txt or copy at
 # http://www.boost.org/LICENSE_1_0.txt)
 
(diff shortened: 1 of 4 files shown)


src/tools/features/coverage-feature.jam
```diff
@@ -1,12 +1,13 @@
...
-# What kind of coverage information to generate.
-
 import feature ;
 
 feature.feature coverage
-    : off all test profile
-    : propagated optional ;
+    :
+    on          # Enable coverage generation for the tool.
+    off         # Disable coverage generation for the tool.
```
grafikrobot: `off` needs to be the default by placing it first on the values.

HDembinski: Right... fixed in the new patch.

PR #482, opened 2019-09-17
Stick to C++98 features for now

### LINKED ISSUE #473: No longer builds in ibm and sun
Both xlc and sunstudio use different flags for version. '--version' doesn't work there.

tools/build/src/engine/build.sh essentially breaks in the initial bootstrapping, before even trying to build b2.

for Sun, it should be CC -V
For ibm, it should be xlC_r -qversion

I can submit a pull request for this, but I'd like to know if the (intentional, as documented in the release note) move to C++11 takes into account that it essentially removes boost support for 98/03. Even if I patch this flag, build will still fail in those compilers.

tanzislam:
Fixes #473. Tested with the "borland" toolchain, which still uses `bcc32.exe`.


tanzislam:
Oops, the perils of developing directly on `develop`. :slight_smile: Re-raised as #485.

PR #494, opened 2019-10-24
update stdlib feature

grisumbras:
Adds gnu11 (libstdc++ with new ABI) and libc++ values to stdlib feature.
The value gnu was already in use by sun toolset, so this commit sets it as
libstdc++ with old ABI. Adds support for that feature to gcc, clang-linux and
clang-darwin toolsets. Refactors sun toolset to use stdlib feature via toolset
flags.


grisumbras:
Sure, I will do it this evening.

PR #495, opened 2019-10-24
add LTO support

grisumbras:
Adds lto feature with 2 values: off (the default) and on. The feature
enables link-time optimizations. Also adds support for the feature to
gcc, clang and msvc toolsets.

### DIFF
--- a/src/tools/clang-darwin.jam
+++ b/src/tools/clang-darwin.jam
@@ -16,9 +16,9 @@ import generators ;
 
 feature.extend-subfeature toolset clang : platform : darwin ;
 
-toolset.inherit-generators clang-darwin 
-  <toolset>clang <toolset-clang:platform>darwin 
-  : gcc 
+toolset.inherit-generators clang-darwin
+  <toolset>clang <toolset-clang:platform>darwin
+  : gcc
   # Don't inherit PCH generators. They were not tested, and probably
   # don't work for this compiler.
   : gcc.mingw.link gcc.mingw.link.dll gcc.compile.c.pch gcc.compile.c++.pch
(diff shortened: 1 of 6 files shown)


src/tools/clang-darwin.jam
```diff
@@ -90,6 +90,10 @@ toolset.flags clang-darwin.compile OPTIONS <inlining>full : -Wno-inline ;
 # shouldn't this be handled by the common gcc?
 toolset.flags clang-darwin.compile OPTIONS <flags> ;
 
+# LTO
+toolset.flags clang-darwin.compile OPTIONS <lto>on : -flto=thin ;
+toolset.flags clang-darwin.link OPTIONS <lto>on : -flto=thin ;
```
grafikrobot: Why use thin instead of using the compiler default? Either we leave it up to the compiler, or we should add a second feature to control the kind of LTO to use.

src/tools/features/lto-feature.jam
```diff
@@ -0,0 +1,12 @@
+# Copyright 2019 Dmitry Arkhipov
+# Distributed under the Boost Software License, Version 1.0.
+# (See accompanying file LICENSE_1_0.txt or copy at
+# http://www.boost.org/LICENSE_1_0.txt)
+
+# TODO: Documentation.
```
grafikrobot: Definitely need some docs here ;-)

src/tools/clang-darwin.jam
grisumbras: I added lto-mode subfeature that allows to choose between thin and full. Also, made `-flto` the default.

src/tools/features/lto-feature.jam
grisumbras: Added docs.

grisumbras:
Ok, now I think the feature is ready.

PR #503, opened 2019-11-06
implement configurable installation prefixes that use features

grisumbras:
This PR adds rule `add-install-dir` that registers a named installation directory that works similar to how those are [treated in Autotools](https://www.gnu.org/prep/standards/html_node/Directory-Variables.html). Only this one uses features instead of variables. As a result, one can do things like
* request `b2 install-prefix=/usr install-bindir=/bin` to install executables into `/bin` and eveything else into `/usr`.
* use conditional properties so that `b2 address-model=32,64` puts 32 bit libraries into `/usr/lib/i686-pc-linux-gnu`, and 64 bit libraries into `/usr/lib/x86_64-pc-linux-gnu`.

By default every named directory supported by Autotools is added.

Also, feature `<install-package>` is added to control target's package name (used on Windows and for directories like `docdir`). If corresponding property isn't present in property set, the package name is guessed from the project's id or directory name.

Another feature `<staging-prefix>` is added to allow putting installed files into a different directory hierarchy (similar to Make's `destdir`).

Finally, two rules are added that allow getting the path of a named directory and package name for a given property set. This allows generating configuration files based on active properties.

All user-facing rules and features are documented.


grisumbras:
I do, though I haven't had much time recently.

grisumbras:
This PR adds rule `add-install-dir` that registers a named installation directory that works similar to how those are [treated in Autotools](https://www.gnu.org/prep/standards/html_node/Directory-Variables.html). Only this one uses features instead of variables. As a result, one can
* request `b2 install-bindir=/bin` to install executables into `/bin`.
* use conditional properties so that `b2 address-model=32,64` puts 32 bit libraries into `/usr/lib/i686-pc-linux-gnu`, and 64 bit libraries into `/usr/lib/x86_64-pc-linux-gnu`.

By default every named directory supported by Autotools is added.

Also, feature `<install-package> is added to control target's package name (used on Windows and for directories like `docdir`). If corresponding property isn't present in property set, the package name is guessed from the project's id or directory name.

Another feature `<staging-prefix> is added to allow putting installed files into a different directory hierarchy (similar to Make's `destdir`).

Finally, two rules that allow getting the path of a named directory and package name for a given property set. This allows generating configuration files based on active properties.

All user-facing rules and features are documented.

src/tools/stage.jam
```diff
@@ -29,6 +37,162 @@ feature.feature <so-version>           :        : free incidental ;
 feature.feature <install-no-version-symlinks> : on : optional incidental ;
 
 
+#| tag::features-doc[]
```
grafikrobot, line 40: Use a tag that reflects the feature name. So that it's easy to tell in the include what it is.

src/tools/stage.jam
```diff
@@ -29,6 +37,162 @@ feature.feature <so-version>           :        : free incidental ;
...
+
+|# # end::doc[]
+
+rule install-dir-names ( )
+{
+    return $(.dirs) ;
+}
+
+
+#| tag::features-doc[]
```
grafikrobot, line 112: Feature specific tag please.

src/tools/stage.jam
```diff
@@ -29,6 +37,162 @@ feature.feature <so-version>           :        : free incidental ;
...
+add-install-dir htmldir : "" : docdir ;
+add-install-dir dvidir : "" : docdir ;
+add-install-dir pdfdir : "" : docdir ;
+add-install-dir psdir : "" : docdir ;
+add-install-dir lispdir : emacs/site-lisp : datarootdir ;
+add-install-dir localedir : locale : datarootdir ;
+add-install-dir mandir : man : datarootdir ;
+
+
+#| tag::features2-doc[]
```
grafikrobot, line 173: Feature specific tag please.

PR #507, opened 2019-11-15
Allow replacing ar flags

hyc:
We need to be able to add "D" option for deterministic builds.


hyc:
The more straightforward approach would be to have ARFLAGS completely replace the default "rc". But one way or another we need a way to change the default.

PR #514, opened 2019-12-16
Add png16 and libpng16 to library names in libpng.jam

mloskot:
On Windows, this allows to successfully find the library installed using vcpkg which deploys the library named as `libpng16.lib` and not `png16.lib` or `png.lib`.

### DIFF
--- a/src/tools/libpng.jam
+++ b/src/tools/libpng.jam
@@ -24,6 +24,10 @@ import property ;
 import property-set ;
 
 header = png.h ;
+
+# On Windows, binary distributions of libpng and package managers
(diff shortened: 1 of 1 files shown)


src/tools/libpng.jam
```diff
@@ -16,15 +16,15 @@ import ac ;
...
+import targets ;
 import path ;
 import modules ;
 import indirect ;
 import property ;
 import property-set ;
 
 header = png.h ;
-names = png ;
+names = libpng libpng16 png png16 ;
```
grafikrobot: Could you add a comment as to why we need those multiple names.

mloskot: Good point. Comment added.

PR #517, opened 2019-12-18
define OSPLAT: MIPS64/MIPS32 instead of MIPS

wzssyqa:
context need to define abi as o32 for MIPS32, while n64 for MIPS64.
we need a way to know about it.

### DIFF
--- a/src/engine/jam.h
+++ b/src/engine/jam.h
@@ -412,7 +412,11 @@
 #endif
 
 #ifdef __mips__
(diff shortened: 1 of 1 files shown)


olk:
Does it pass the unit tests?

wzssyqa:
> 
> 
> Does it pass the unit tests?

with this group of patch, I run unittest of context on mips64el, it works.
In fact some test fails due to lack of -fsplit-stack support of the gcc for mips.

PR #523, opened 2019-12-29
bugfix: saxonhe_dir actions now conditional on build host OS

madmongo1:
When I write the saxionhe_dir actions clause I did not realise that these actions are copied to the command line without interpretation with respect to the build host OS.

As a result, I fixed beast documentation building for linux, but broke it for windows.

This change fixes it for both.

### DIFF
--- a/src/tools/saxonhe.jam
+++ b/src/tools/saxonhe.jam
@@ -7,6 +7,7 @@
 #
 
 import common ;
+import os ;
 
 rule init ( saxonhe_jar ? : java_exe ? )
 {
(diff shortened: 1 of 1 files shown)


vinniefalco:
Suggested commit message:

> saxonhe_dir works on Windows

src/tools/saxonhe.jam
```diff
@@ -34,8 +35,19 @@ actions saxonhe
 # execute the saxonhe jar file passing directories as inputs and outputs.
 # saxonhe requires that the output directory already exists
 #
-actions saxonhe_dir
+if [ os.name ] = NT
```
grafikrobot: Use `[ os.on-windows ]`.

src/tools/saxonhe.jam
```diff
@@ -34,8 +35,19 @@ actions saxonhe
...
 {
-    mkdir -p "$(<)"
-    "$(.java_exe)" -jar "$(.saxonhe_jar)" -o:"$(<)" -s:"$(>[1])" -xsl:"$(>[2])"
+    actions saxonhe_dir
+    {
+        MKDIR "$(<)"
```
grafikrobot: You might need to make this conditional on dir exists as per common.mkdir action: https://github.com/boostorg/build/blob/develop/src/tools/common.jam#L725

src/tools/saxonhe.jam
madmongo1: Thanks for that. As you've probably guessed, I'm a bjam noob. Vinnie and I just copied what we saw elsewhere.

src/tools/saxonhe.jam
madmongo1: I don't suppose there's a way to call an action from an action, in order to make the jamfiles more DRY?

src/tools/saxonhe.jam
vinniefalco: @madmongo1 we should change it in all my Jamfiles (e.g. Beast, JSON, URL)

grafikrobot: @madmongo1 no problem :-) I consider things like this "expert" level since they aren't easy to find (lack of visible documentation).

src/tools/saxonhe.jam
grafikrobot: Nothing direct at the moment. It could be done with some more "expert" level rule calling actions code. But this is definitely a feature that would be good to have :-)

vinniefalco:
doesn't this need to go into _master_ to fix the broken superproject build?

PR #531, opened 2020-01-29
Add -std=c++11 flag because sysinfo uses c++11 features

### LINKED ISSUE #530: sun toolset needs to pass -std=c++11 to compiler
As sysinfo uses c++11 features build.sh needs to pass -std=c++11 to compiler. Otherwise build fails

vahtis:
Fixes #530 

src/engine/build.sh
Added -std=c++11 flag to compile flags

### DIFF
--- a/src/engine/build.sh
+++ b/src/engine/build.sh
@@ -314,7 +314,7 @@ case $B2_TOOLSET in
             export PATH
         fi
         B2_CXX="${CXX}"
-        B2_CXXFLAGS_RELEASE="-xO4 -s"
+        B2_CXXFLAGS_RELEASE="-xO4 -s -std=c++11"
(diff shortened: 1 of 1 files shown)


src/engine/build.sh
```diff
@@ -314,7 +314,7 @@ case $B2_TOOLSET in
             export PATH
         fi
         B2_CXX="${CXX}"
-        B2_CXXFLAGS_RELEASE="-xO4 -s"
+        B2_CXXFLAGS_RELEASE="-xO4 -s -std=c++11"
```
grafikrobot: The `B2_CXXFLAGS_RELEASE` is for additional optimization flags. You need to move the new option to the `B2_CXX` var.

vahtis: Done

PR #536, opened 2020-02-05
import errors.error in targets.resolve-reference.

jehelset:
got an unknown-error instead of error-message without the import.


src/build/targets.jam
```diff
@@ -887,6 +887,7 @@ rule resolve-reference ( target-reference : project )
     local id = $(split[1]) ;
     if ! $(split) || ! $(id)
     {
+        IMPORT errors : error : : error ;
```
grafikrobot: Should be something like this instead:
```
import errors ;
errors.error "..." ;
```

PR #538, opened 2020-02-20
Fix non-zero exit code on Jam parsing errors

mloskot:
Prevent resetting the non-Zero status returned from yyanyerrors on any parsing errors.

-----

The bug was reported on #boost Slack

------

A basic verification of the fix:

1. `Jamfile` with a broken syntax

image

2. Before the fix, `b2` returns Zero

image

3. After the fix, `b2` returns non-Zero

image

### DIFF
--- a/src/engine/jam.cpp
+++ b/src/engine/jam.cpp
@@ -275,6 +275,7 @@ int main( int argc, char * * argv, char * * arg_environ )
     b2::system_info sys_info;
 
     saved_argv0 = argv[ 0 ];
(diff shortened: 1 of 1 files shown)


mloskot:
My initial attempt to fix this with unconditional ` last_update_now_status = status;` was incorrect.

Since, AFAIU, `last_update_now_status` may be set before `status` is set in line `status = yyanyerrors();`, unconditionally setting the last status with `status` had undesired effects.

I have updated the PR according to my current understanding.

PR #556, opened 2020-03-25
The dinkumware library used by the Embarcadero clang-based compilers …

eldiener:
…does not work with rtti turned off. I have reported this to Embarcadero as a potential bug but I strongly suspect that this is by design.


eldiener:
I have no idea what has failed even when looking at the "Details". All the tests run show that they have passed.

PR #558, opened 2020-03-28
`<force-include>` feature

### LINKED ISSUE #387: Forced include feature
I did not find forced include feature which I wish to use with PCH (adding PCH with automatic including the header without spoiling out `#include` in every file). All major compilers support it:

Compiler | Flag | Since version
-------- | ---- | -------------
MSVC     | [`/FI`](https://docs.microsoft.com/en-us/cpp/build/reference/fi-name-forced-include-file?view=vs-2017) | [VC4 had it](http://gunkies.org/wiki/Visual_C%2B%2B_4.0)
GCC      | [`-include`](https://gcc.gnu.org/onlinedocs/gcc/Preprocessor-Options.html#index-include) | [GCC 3.4 had it](https://gcc.gnu.org/onlinedocs/gcc-3.4.0/gcc/Preprocessor-Options.html#Preprocessor-Options)
Clang    | [`/FI` `-include`](https://clang.llvm.org/docs/UsersManual.html) | [Clang 3.1 had `-include`](https://releases.llvm.org/3.1/tools/clang/docs/UsersManual.html)

Kojoley:
Closes #387

Some nitpicks:

1. `-include\ "$(FORCE_INCLUDES)"` does not work because despite of [documentation](https://boostorg.github.io/build/manual/develop/index.html#jam.language.lexical) a backslash does not escape a whitespace, used `-include"$(FORCE_INCLUDES)"` instead (and it surprisingly worked out).

2. I did not add this to MSVC MIDL/RC or GCC Fortran as I am not sure if they need it.


Kojoley:
Any comments?

src/tools/features/force-include-feature.jam
```diff
@@ -0,0 +1,19 @@
+# Copyright 2020 Nikita Kniazev
+# Distributed under the Boost Software License, Version 1.0.
+# (See accompanying file LICENSE_1_0.txt or copy at
+# http://www.boost.org/LICENSE_1_0.txt)
+
+import feature ;
+
+#| tag::doc[]
```
grafikrobot, line 8: For the docs you also need to add to the `reference.adoc` file.

src/tools/features/force-include-feature.jam
```diff
@@ -0,0 +1,19 @@
...
+
+[[bbv2.builtin.features.force-include]]`force-include`::
+Specifies an include path that has to be included in a way like if
+`#include "file"` appeared as the first line of the primary source file.
```
grafikrobot, line 12: Mention what happens when multiple ones are specified.

src/tools/features/force-include-feature.jam
```diff
@@ -0,0 +1,19 @@
...
+
+|# # end::doc[]
+
+feature.feature "force-include"
+    :
+    : free #order-sensitive
```
grafikrobot, line 18: Remove the commented out part.

test/feature_force_include.py
```diff
@@ -0,0 +1,39 @@
...
+
+# Tests the force-include feature
+
+import BoostBuild
+
+t = BoostBuild.Tester(use_test_config=False)
+
+t.write("Jamroot.jam", """
+obj test-cpp : test.cpp : <force-include>test.h ;
+obj test-c : test.c : <force-include>test.h ;
```
grafikrobot, line 16: Could you also test when specifying multiple `<force-include>`.

src/tools/features/force-include-feature.jam
Kojoley: Does it even work? I copied it from `<include>` and I remember some discussions about making it working but as far as I know it is still not the case.

test/feature_force_include.py
Kojoley: Would it be ok to test only multiple or I need to create another test?

src/tools/features/force-include-feature.jam
grafikrobot: I meant literally delete the `#order-sesitive` text. No point in having it there if it's not doing anything.

test/feature_force_include.py
grafikrobot: Sure, that's fine.

Kojoley:
> Not sure I like the "force-include" name. Other names I thought of: pre-include and first-include.

I took the name from the MSVC. Maybe I like `pre-include` more, but it creates a false sense of it being before `include` while they are distinct things.

src/tools/features/force-include-feature.jam
Kojoley: Ok...

src/tools/features/force-include-feature.jam
Kojoley: Done

PR #559, opened 2020-03-28
Embarcadero C++ clang-based compilers only support C++17, so cxxstd i…

eldiener:
…s turned off for all ISO levels.


eldiener:
Any objections to merging this PR ?

PR #564, opened 2020-04-07
Improved Embarcadero generated command lines by using response files. Corrected system library paths.

eldiener:


eldiener:
Something is wrong with your appveyor setup.

Issue #261, opened 2017-11-12
Alternative output format with color


jpcima:
Hi. Can we discuss the possibility of improving the format of boost.build's console output?
I have created my own [patch](https://github.com/jpcima/build/commit/3706e07b53c523deef2e30df986b6ba3bd629856) in order to to make the output of my builds more visually pleasing.
By default, the build output occupies a lot of screen space and is not easily distinguished from the output of the compiler. I took some screen captures for [comparison](https://gist.github.com/jpcima/1e35747f78c5d7e0642c5fb4e1e35eeb).
The code is for Windows and ANSI terminals and does not disrupt pipe output. I listen to your comments and suggestions.

swatanabe:
There should be an option to control this explicitly.  Also, it would be better to use descriptive names like TC_ACTION_NAME instead of naming a specific color.

mloskot:
There was a brief discussion on such feature here https://svn.boost.org/trac10/ticket/3508 followed-up on the mailing list, see https://lists.boost.org/boost-build/2012/01/25582.php

Wouldn't the colourful messages implemented as part of the `ECHO`  be more B2 canonical? 
What is the right way to appraoch this

PR #573, opened 2020-04-21
WIP: Add colors support to echo rule and actions

mloskot:
This is my first stab at native support for colourful output from the built-in `ECHO` rule and other sources like actions (name, target, command).

It's been long time since I brainstormed this idea, https://svn.boost.org/trac10/ticket/3508 and https://lists.boost.org/boost-build/2012/01/25582.php, then implemented it as Perl script filter https://github.com/mloskot/colorbb, but lately, I've started looking into B2 source code more. Inspired by @jpcima's #261 and @swatanabe's https://github.com/boostorg/build/issues/261#issuecomment-352288362 , I picked it up and _experimented_ a bit.

It is not a feature-complete optimal implementation, but an experimental proposal to collect feedback about direction of this feature. It is working and usable though, on both, Windows and Unix. It uses basic set of [8/16 ANSI colors](https://misc.flogisoft.com/bash/tip_colors_and_formatting).

The outout is only coloured if requested specifying the `-c` option in the command line.

### Examples: Windows

I used ConEmu w/ Monokai color scheme on Windows.

Initial examples of the colourful output from the `ECHO`, see the `example/colors/Jamfile`, built with `./b2 -c`

image

image

and built with `./b2 -c -d+2`

image

image

### Examples: Linux

image

image

------

(The CMake files bundled in the first commit are just for temporary convenience, for myself and others who may find it useful for testing of this PR. If this PR ever becomes ready to merge, all the CMake cruft will be removed.)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,7 @@
 *.code-workspace
 *.py[co]
+/.idea
+/.vs
 /.vscode
 /doc/bin
 /doc/html
@@ -8,7 +10,9 @@
 /b2.exe
 /bjam.exe
 /.build
+/_build*
 /bootstrap.log
+/cmake-build-*
 /test/test_results.txt
 /project-config.jam
 /src/engine/b2
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,0 +1,76 @@
+cmake_minimum_required(VERSION 3.10)
+
+project(boost_build LANGUAGES CXX)
+
+set(CMAKE_CXX_STANDARD 11 CACHE STRING "C++ standard version to use (default is 11)")
+set(CMAKE_CXX_STANDARD_REQUIRED ON)
+set(CMAKE_CXX_EXTENSIONS OFF)
+
+add_executable(b2)
+
+target_sources(b2
+  PRIVATE
+    ${CMAKE_CURRENT_SOURCE_DIR}/src/engine/builtins.cpp
+    ${CMAKE_CURRENT_SOURCE_DIR}/src/engine/class.cpp
+    ${CMAKE_CURRENT_SOURCE_DIR}/src/engine/color.cpp
+    ${CMAKE_CURRENT_SOURCE_DIR}/src/engine/command.cpp
+    ${CMAKE_CURRENT_SOURCE_DIR}/src/engine/compile.cpp
+    ${CMAKE_CURRENT_SOURCE_DIR}/src/engine/constants.cpp
+    ${CMAKE_CURRENT_SOURCE_DIR}/src/engine/cwd.cpp
+    ${CMAKE_CURRENT_SOURCE_DIR}/src/engine/debug.cpp
+    ${CMAKE_CURRENT_SOURCE_DIR}/src/engine/debugger.cpp
+    ${CMAKE_CURRENT_SOURCE_DIR}/src/engine/execcmd.cpp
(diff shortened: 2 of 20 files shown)


eldiener:
Personally I always 'tee' the b2 output to a file, which I then view in my editor, JEdit. But maybe others will find your "colorful" b2 output helpful.

mloskot:
@eldiener 
> Personally I always 'tee' the b2 output to a file (...) maybe others will find your "colorful" b2 output helpful.

I guess, you are not finding colours from GCC or clang helpful either. No worry, I'm proposing opt-in colouring.

FYI, use of `tee` can be replaced with `-o` option: `b2 -olog.txt` which tells `b2` to mirror in `log.txt` file everything it sends to `stdout` and `stderr`.

eldiener:
Thanks for the information about the b2 -o option.

HDembinski:
I love this very much.

mloskot:
@grafikrobot I have two issues which I'd like to reach your opinion about

### What is preferred `ECHO` argument format for colors?

I [brainstormed variants on the Slack](https://cpplang.slack.com/archives/C27KZLB0X/p1587069125476300).

Basically, is it better to have single dedicated argument for all colors (and their hyphen-delimited attributes):

```
ECHO "red on default" : red ;
ECHO "red on blue" : red blue ;
```

or better is to split the foreground and background into two separate formal arguments:

```
ECHO "red on blue" : red : blue ;
```

### Do we want bold text too or just colors (normal and light)?

I'm going to clean up handling of the internal flags for colors to better support light colors on **Unix** terminals using dedicated escape codes.

For example, currently, red is `\e[31` and additional attribute `\e[1` will make it brighter but also text becomes bold. This needs to become `\e[31`  for red , `\e[91` for light red, etc. to allow this without affecting text:

```
ECHO "red on blue" : red-light blue ;
ECHO "red on blue" : red blue-light;
``` 

Then, the `\e[1`code could control bold text separately, for example:

```
ECHO "red on blue" : red-bold ;
ECHO "red on blue" : red-light-bold blue ;
``` 

Question is, is bold attribute really useful or needed?

Kojoley:
I know I could try it myself but it is easier to ask. Does this paint compile fail test output (fail-as-expected) in red?

mloskot:
@Kojoley it should paint it green, as it's successful result. See the screenshots.

mloskot:
@grafikrobot Excellent, the modular arguments handling is what I also prefer. I initially went for squeezing all colors into single argument in case `ECHO` gets other arguments in future. Then, it may be better to pass all colors stuff as single one.

I will make necessary changes to this PR soon.

chambm:
I'd love to have colors for my Windows builds. I actually tried to do this a few years ago but failed, so I'm glad you got it working. Does it work in cmd?

mloskot:
@chambm 
> Does it work in cmd? 

Yes, it works in the Windows Command Prompt too.

> Can we get it to colorize lines conditionally like coloring warnings, errors, and notes distinctly even if they're from the same compile command?

In this first stab, I focused on the basic framework, but I think I've achieve a basic implementation of something like that.
For fine grained colouring,  I'll have to dig it deeper into the engine implementation, what I am going to do.

FYI, this is PR is not abandoned. I had to put it on backburner as unexpected duties due to COVID19 have been eating all my free time.

PR #575, opened 2020-04-23
Use /Zc:inline and /Zc:throwingNew flags

Kojoley:
Size differences of `b2 toolset=msvc-14.2 variant=release link=shared,static`

flags              | bin.v2/libs   | diff | lib         | diff | dll        | diff
-------------------|--------------:|-----:|------------:|-----:|-----------:|-----:
no                 | 1 268 748 304 |      | 393 665 694 |      | 11 630 592 |  
inline	           |   757 974 914 | -40% | 173 383 766 | -56% | 11 615 744 | -0.1%
inline+throwingNew |   757 909 662 |  -0% | 173 444 188 |  +0% | 11 590 144 | -0.2%

Timings were 765s/732s/757s

Links to docs:
https://docs.microsoft.com/en-us/cpp/build/reference/zc-inline-remove-unreferenced-comdat?view=vs-2019
https://docs.microsoft.com/en-us/cpp/build/reference/zc-throwingnew-assume-operator-new-throws?view=vs-2019

### DIFF
--- a/src/tools/msvc.jam
+++ b/src/tools/msvc.jam
@@ -454,6 +454,14 @@ rule configure-version-specific ( toolset : version : conditions )
         }
     }
 
+    # 12.0 (VS2013 Update 2) introduced /Zc:inline opt-in standard conformance
+    # compiler flag that also similar to linker /opt:ref removes unreferenced
+    # variables and functions that have internal linkage
+    if ! [ MATCH "^(1[01]|\d)\\." : $(version) ]
+    {
+        toolset.flags $(toolset).compile CFLAGS $(conditions) : "/Zc:inline" ;
+    }
+
     #
     # Processor-specific optimization.
     #
--- a/src/tools/msvc.jam
+++ b/src/tools/msvc.jam
@@ -462,6 +462,13 @@ rule configure-version-specific ( toolset : version : conditions )
         toolset.flags $(toolset).compile CFLAGS $(conditions) : "/Zc:inline" ;
     }
 
+    # 14.0 introduced /Zc:throwingNew opt-in flag that disables a workaround
+    # for not throwing operator new in VC up to 6.0
+    if ! [ MATCH "^(1[0-3]|\d)\\." : $(version) ]
+    {
+        toolset.flags $(toolset).compile CFLAGS $(conditions) : "/Zc:throwingNew" ;
+    }
+
     #
     # Processor-specific optimization.
     #
--- a/src/tools/msvc.jam
+++ b/src/tools/msvc.jam
@@ -229,6 +229,7 @@ import set ;
 import toolset ;
 import type ;
 import virtual-target ;
+import version ;
 
 
 type.register MANIFEST : manifest ;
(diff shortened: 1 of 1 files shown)


Kojoley:
Ping

This is better to be merged sooner to give enough time for any potential problem to show up and be fixed (even while I am not expecting to see one).

src/tools/msvc.jam
```diff
@@ -454,6 +454,21 @@ rule configure-version-specific ( toolset : version : conditions )
         }
     }
 
+    # 12.0 (VS2013 Update 2) introduced /Zc:inline opt-in standard conformance
+    # compiler flag that also similar to linker /opt:ref removes unreferenced
+    # variables and functions that have internal linkage
+    if ! [ MATCH "^(1[01]|\d)\\." : $(version) ]
+    {
+        toolset.flags $(toolset).compile CFLAGS $(conditions) : "/Zc:inline" ;
```
grafikrobot, line 462: I would prefer that `/Zc:inline` be either tied to an existing feature or a new feature added for it. It would be surprising to users to add it unconditionally for users. Especially if they specifically want the opposing functionality.

src/tools/msvc.jam
```diff
@@ -454,6 +454,21 @@ rule configure-version-specific ( toolset : version : conditions )
...
+    }
+
+    # 14.0 introduced /Zc:throwingNew opt-in flag that disables a workaround
+    # for not throwing operator new in VC up to 6.0
+    if ! [ MATCH "^(1[0-3]|\d)\\." : $(version) ]
+    {
+        toolset.flags $(toolset).compile CFLAGS $(conditions) : "/Zc:throwingNew" ;
```
grafikrobot, line 469: I can see how this could be the default. But like above I would prefer that it there is a way to control if it's added or not. The only current feature to tie it to would be the language level. But that also doesn't seem correct. As only users know what behavior they want.

src/tools/msvc.jam
Kojoley: We can tie this one to `<visibility>hidden`, however I do not understand why `/Zc:forScope` and `/Zc:wchar_t` are applied unconditionally while these you are asking to go under a feature when it is a misfeature and a compiler bug, from my point of view it should be a compiler default, anyway users could `<cxxflags>/Zc:inline-` if they want to enable this compiler bug back.

src/tools/msvc.jam
Kojoley: The same thing as above, but for this one I do not see any potential existing feature to integrate into

src/tools/msvc.jam
grafikrobot: The `forScope` and `wchar_t` are just old. They where added 16 years ago. Not sure we would do those the same way now either. But not sure we can do different now. And for those not sure it matters as no one uses msvc-7 any longer (or at least I hope they don't).

Kojoley: And you completely ignored `<cxxflags>/Zc:inline-` part :-)
I do not know how to proceed, adding 18 features for every conformance flag looks like a bad idea.

grafikrobot: Sorry.. brain is fragmented from trying to work at the same time :-) I do agree that adding a bunch of features for a bunch of flags is not desirable. Haven't thought of an alternative though. Okay, I give up :-)

PR #581, opened 2020-04-25
fix: link statically against boost-python on linux

### LINKED ISSUE #580: cannot link statically against boost-python on linux
i cannot link statically against boost-python on linux, because it links against pthread statically in that case, which is not allowed

joriscarrier:
closed #580

### DIFF
--- a/src/tools/python.jam
+++ b/src/tools/python.jam
@@ -51,7 +51,10 @@ lib dl ;
 # pass this to linker explicitly.
 lib util ;
 # Python uses pthread symbols.
-lib pthread ;
+lib pthread :
+    : <target-os>linux:<link>shared
+    ;
+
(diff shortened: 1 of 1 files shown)


src/tools/python.jam
```diff
@@ -51,7 +51,20 @@ lib dl ;
 # pass this to linker explicitly.
 lib util ;
 # Python uses pthread symbols.
-lib pthread ;
+lib pthread :
+    : <conditional>@pthread_linking
```
grafikrobot, line 55: This all looks equivalent to adding `<target-os>linux:<link>shared` instead of using a full conditional rule. Also, does this apply to other systems that use pthread? I.e. should pthread always be shared link?

joriscarrier: no problem with non-linux systems, pthread can be link statically

PR #582, opened 2020-04-27
fix(configure): add relevant feature to build a simple check program

ledocc:
when checking if a lib is available by linking an executable, we have to
build with the same variant/link/define that the final build.

### DIFF
--- a/src/build/configure.jam
+++ b/src/build/configure.jam
@@ -41,7 +41,7 @@ feature.feature configure : : composite optional ;
 # relevant features are also considered relevant.
 #
 feature.compose <configure> :
-    <target-os> <toolset> <address-model> <architecture> <cxxstd> ;
+    <target-os> <toolset> <address-model> <architecture> <cxxstd> <variant> <link> <runtime-link> <define> ;
 
 
 rule log-summary ( )


ledocc:
up

ledocc:
@swatanabe, can you take a look to this PR please ?

swatanabe:
AMDG

This should only include propagated features.  It should not
include <define>.

I'm wondering, though, whether it would be better to drop the hard-coded
list entirely.  No matter what set of properties we pick they'll be
wrong for some uses.  We can generate the cache key using
[ $(ps).base-relevant ] or similar.

In Christ,
Steven Watanabe

ledocc:
I do this patch when I try to build boost.locale with ICU support on windows,
My first problem was boost.locale test if ICU is available with a simple test program "has_icu".
But boost.build use the "debug" variant even if I build boost in "release" mode. But MSVC does not allow
to link debug with release lib/executable. 
So I add "variant" in the list of "relevant" feature.
Same kind of issue when I build with static build of ICU that require a preprocessor DEFINE variable. 
Then I add  "define" in the list of "relevant" feature.
Same story for link and runtime-link

```
# The configure feature allows external definition of what features are
# relevant for doing configuration builds. One can add additional relevant
# features by using:
#
#   import feature ;
#   import configure ;
#   feature.compose <configure> : <threading> ;
#
```
I try this but no success, so I change it directly in "src/build/configure.jam" ...

Issue #555, opened 2020-03-25
build boost1.72 on windows10 with msvc14.2(vs2019) failed


liqi198786:
it will always occur the error as below:
"failed to write output file 'build\boost\architecture\msvc-14.2\debug\threading-multi\32.obj.rsp'!"
i searched the error on the net, some people suggest adding some options such as "cxxflags=/FS" "pch=off", but these didn't work for my case.
and here is all my cmd:
"b2 -j32^
    headers^
    --layout=versioned^
    --build-dir=.\build^
    --with-system^
    --with-date_time^
    --with-filesystem^
    --with-python^
    --project-config=project-config.jam^
    -d 1^
    architecture=x86^
    address-model=64^
    toolset=msvc-14.2^
    variant=release^
    link=static^
    threading=multi^
    runtime-link=shared^
    --prefix="C:\Users\szx3gg\Downloads\boost_1_72_0"^
    --libdir="C:\Users\szx3gg\Downloads\boost_1_72_0\lib"^
    --includedir="C:\Users\szx3gg\Downloads\boost_1_72_0"^
    install
    "
i also have tried to test remove some options in the cmdline or just keep them like "threading=multi^" , but still failed with the error.

grafikrobot:
Try adding either `--abbreviated-paths` or `--hash` options (https://boostorg.github.io/build/manual/master/index.html#bbv2.reference.buildprocess.targetpath). If that works it means your setup is creating long paths. Depending on which version of Windows you have it might be possible to enable long path support in the OS.

liqi198786:
> Try adding either `--abbreviated-paths` or `--hash` options (https://boostorg.github.io/build/manual/master/index.html#bbv2.reference.buildprocess.targetpath). If that works it means your setup is creating long paths. Depending on which version of Windows you have it might be possible to enable long path support in the OS.

hi @grafikrobot , thanks for your answer, but it seems that doesn't work for me either. i now really wonder what .rsp file is, and why the .obj file needs to be transformed or something like to the .rsp file format.

grafikrobot:
The RSP files are "response" files. They contain the arguments tot he compiler for building that particular source file. They get used to work around compiler command invocation argument length limits.

Not sure what else, other than the path length, could be going wrong here :-(

liqi198786:
hi @grafikrobot , i also tried the simple example in the b2 folder to build a project and built it with b2.exe. the project is as below:
[...]

PR #583, opened 2020-04-28
Print fopen fail reason

Kojoley:
Should help investigate #555

### DIFF
--- a/src/engine/execvms.cpp
+++ b/src/engine/execvms.cpp
@@ -169,7 +169,8 @@ void exec_cmd
         /* Open tempfile. */
         if ( !( f = fopen( commandbuf + 1, "w" ) ) )
         {
-            printf( "can't open cmd_string file\n" );
+            err_printf( "[errno %d] failed to wite cmd_string file '%s': %s",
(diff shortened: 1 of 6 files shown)


Kojoley:
```
Actual STDERR {{{
[errno 2] failed to scan file 'dir\bin\hello1.cpp': No such file or directory
}}}
```

I tried to understand whether this is intentional or a bug but I have no idea. Why b2 tries to find a source file in a build directory?

mloskot:
Smells like [Perforce Jam](https://www.perforce.com/documentation/jam-documentation) legacy.

Kojoley:
Ok, it looks for it because it should be created by `make` rule:
```
t.write("dir/jamfile.jam", """\
import common ;
make hello1.cpp : hello.cpp : common.copy ;
""")
```
I see a message that `hello1.cpp` is copied and found `hello1.obj` in build directory but there is no `hello1.cpp`, it seems to be removed by b2 and afterwards it tries to scan its content, is not it a bug?

Kojoley:
It is because the test is run with `-n Don't execute the commands, only print them` flag

PR #592, opened 2020-05-06
Add cxxstd value 20 for C++20

Lastique:
The -std=c++20 is supported by clang 10.

### DIFF
--- a/src/tools/clang.jam
+++ b/src/tools/clang.jam
@@ -55,7 +55,10 @@ rule init-cxxstd-flags ( toolset : condition * : version )
     local dialects = [ feature.values <cxxstd-dialect> ] ;
     dialects = [ set.difference $(dialects) : gnu iso ] ;
     local std ;
-    if [ version-ge $(version) : 3.5 ] { std = 1z ; }
+    if [ version-ge $(version) : 10.0 ] { std = 20 ; }
+    else if [ version-ge $(version) : 6.0 ] { std = 2a ; }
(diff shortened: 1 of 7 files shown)


Lastique:
If possible, please merge it soon as I've already modified CI settings in a few submodules to use "20", not knowing Boost.Build doesn't support it yet.

Lastique:
C++20 is practically finalized.

Re toolset changes, do you mean to update "latest"? The "20" value itself doesn't seem to need any changes in the toolsets.

Lastique:
I've updated `cxxstd=latest` for clang. For gcc it is already set for 2a for gcc 8 and newer.

Lastique:
> But we don't change it to "2a" as that's not a final release. For clang-10 it would mean changing latest == 20.

I followed the existing practice re "latest" - set it to the latest C++ version supported by a compiler version, including preliminary tags like "2a". `-std=c++2a` is supported since [clang 6](https://releases.llvm.org/6.0.0/tools/clang/docs/ReleaseNotes.html#id9). I can set it to "20" for clang-10, but I don't really know which compiler versions introduced the final tags like 11, 14 and 17.

> Is there a 2b option in clang-10 now?

I'm not aware of such, I don't think there is.

Lastique:
I've updated "latest" mapping as you requested.

Lastique:
Thanks.

Lastique:
Could you merge this to master, please?

Lastique:
The 1.74 release is imminent. Please, merge to master.

PR #599, opened 2020-05-08
Remove leftover manifest generation early exiting

Kojoley:
It should have been removed in dd3dbcc9b6d4d6322d693f58bb4306930fb19f68


Kojoley:
I also pushed removed unneeded manifest existence check. The commit it was added in d8e98e3abc18dc328a759f6f4bc71741687bc272 says it is fixing the problem that MSVC 8.0 does not produce a manifest file for static libraries, but the code it touches is for shared libraries and executables. And collapsed the branches as they became the same.

PR #601, opened 2020-05-10
Pass /nologo to rc

Kojoley:
Supported at least with Resource Compiler Version 6.1.7600.16385 that comes in Windows SDK v7.0A, the oldest you could obtain nowadays.

### DIFF
--- a/src/tools/clang-win.jam
+++ b/src/tools/clang-win.jam
@@ -175,7 +175,7 @@ rule init ( version ? : command * : options * )
         toolset.flags clang-win.archive .LD $(cond) : $(archiver) /nologo ;
         toolset.flags clang-win.link .MT $(cond) : $(manifest-tool) -nologo ;
         toolset.flags clang-win.compile .MC $(cond) : $(mc-compiler) ;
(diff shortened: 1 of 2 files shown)


src/tools/clang-win.jam
```diff
@@ -175,7 +175,7 @@ rule init ( version ? : command * : options * )
         toolset.flags clang-win.archive .LD $(cond) : $(archiver) /nologo ;
         toolset.flags clang-win.link .MT $(cond) : $(manifest-tool) -nologo ;
         toolset.flags clang-win.compile .MC $(cond) : $(mc-compiler) ;
-        toolset.flags clang-win.compile .RC $(cond) : $(resource-compiler) /nologo ;
```
grafikrobot, line 178: I'm confused.. Is this not already adding `/nologo` to the RC command? It should.. If it's not it's likely a bug.

Kojoley: This is Clang and it adds /nologo, but MSVC and Intel-win does not, so I removed it from Clang and added globally instead.

grafikrobot: Oh, haha.. I missed that.

PR #603, opened 2020-05-11
Fix timestamp_delta_seconds

### LINKED ISSUE #584: Timestamp delta calculation is erroneous
For users interested in getting the execution time of a build step there is a `time` target feature. I have observed the third `clock` line of the output to produce negative numbers. It looks like the fault is in [this code](https://github.com/boostorg/build/blob/develop/src/engine/timestamp.cpp#L227-L230) which converts seconds to microseconds and then adds the result to a nanosecond value.

For example:
> user: [waveidl] 0.500173
> system: [waveidl] 0.069151
> clock: [waveidl] -429.698395

Thanks @mloskot and @pdimov for helping to identify the issue.

Kojoley:
* The multiplier 1000000 is wrong
* Return seconds as function name and description says
* Use `difftime` instead of assuming `time_t` is in seconds

Fixes #584

### DIFF
--- a/src/engine/timestamp.cpp
+++ b/src/engine/timestamp.cpp
@@ -6,6 +6,7 @@
 
 /* This file is ALSO:
  * Copyright 2001-2004 David Abrahams.
+ * Copyright 2020 Nikita Kniazev.
  * Distributed under the Boost Software License, Version 1.0.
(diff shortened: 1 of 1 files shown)


mloskot:
@Kojoley Thanks a lot!

PR #604, opened 2020-05-12
Embed manifest via linker

Kojoley:
The feature is enabled by default only for MSVC 11 and above not to break
derived toolsets.

Unfortunately it cannot be enabled on clang-cl with MSVC linker because of:
```
>clang-cl test.cpp /link /manifest:embed
LINK : fatal error LNK1158: cannot run 'rc.exe'
clang-cl: error: linker command failed with exit code 1158 (use -v to see invocation)
```

Note: `embed-manifest-file` feature was broken before the change and still is
broken under `embed-manifest-via=mt`. The fix seems to be obvious, but I am not
fully understand what happens inside link/link.dll rule to fix it here.

### DIFF
--- a/src/tools/msvc.jam
+++ b/src/tools/msvc.jam
@@ -263,6 +263,16 @@ Input and Output -> Additional Manifest Files.
 
 feature.feature embed-manifest-file : : free dependency ;
 
+#| tag::embed-doc[]
+
+[[bbv2.builtin.features.embed-manifest-via]]`embed-manifest-via`::
+This feature is specific to the `msvc` toolset (see <<Microsoft Visual C++>>),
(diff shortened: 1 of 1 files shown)


src/tools/msvc.jam
```diff
@@ -263,6 +263,16 @@ Input and Output -> Additional Manifest Files.
 
 feature.feature embed-manifest-file : : free dependency ;
 
+#| tag::embed-doc[]
+
+[[bbv2.builtin.features.embed-manifest-via]]`embed-manifest-via`::
+This feature is specific to the `msvc` toolset (see <<Microsoft Visual C++>>),
+and controls whether manifest should be embedded via linker or minifest tool.
```
grafikrobot, line 270: "minifest" ==> "manifest"

Issue #634, opened 2020-07-08
Intel-linux C++11 flag is -std=c++11


aminiussi:
Hi,

With icc/icpc, the c++11 selection flag is the same as g++: -std=c++11.
Right now -xc++ is used in two places of build.sh:

```
[alainm@castor engine]$ git diff
diff --git a/src/engine/build.sh b/src/engine/build.sh
index 41af371..86e1065 100755
--- a/src/engine/build.sh
+++ b/src/engine/build.sh
@@ -81,7 +81,7 @@ test_cxx11 ()
         case $1 in
             gcc) ( ${CXX:=g++} -x c++ -std=c++11 check_cxx11.cpp && rm -f a.out ) 1>/dev/null 2>/dev/null ;;
             intel-darwin) ( ${CXX:=icc} -xc++ check_cxx11.cpp && rm -f a.out ) 1>/dev/null 2>/dev/null ;;
-            intel-linux) ( ${CXX:=icc} -xc++ check_cxx11.cpp && rm -f a.out ) 1>/dev/null 2>/dev/null ;;
+            intel-linux) ( ${CXX:=icc} -std=c++11 check_cxx11.cpp && rm -f a.out ) 1>/dev/null 2>/dev/null ;;
             vacpp) ( ${CXX:=xlC_r} check_cxx11.cpp && rm -f a.out ) 1>/dev/null 2>/dev/null ;;
             xlcpp) ( ${CXX:=xlC_r} check_cxx11.cpp && rm -f a.out ) 1>/dev/null 2>/dev/null ;;
             como) ( ${CXX:=como} check_cxx11.cpp && rm -f a.out ) 1>/dev/null 2>/dev/null ;;
@@ -283,7 +283,7 @@ case $B2_TOOLSET in
             export LD_RUN_PATH
             . ${B2_TOOLSET_ROOT}bin/iccvars.sh $ARCH
         fi
-        B2_CXX="${CXX} -xc++"
+        B2_CXX="${CXX} -std=c++11"
         B2_CXXFLAGS_RELEASE="-O3 -s"
         B2_CXXFLAGS_DEBUG="-O0 -g -p"
     ;;
[alainm@castor engine]$  git status
# On branch master
# Changes not staged for commit:
#   (use "git add <file>..." to update what will be committed)
#   (use "git checkout -- <file>..." to discard changes in working directory)
#
#	modified:   build.sh
#
no changes added to commit (use "git add" and/or "git commit -a")
[alainm@castor engine]$
```

[...]

PR #635, opened 2020-07-08
Bugfix/634 intel linux c++11

aminiussi:
As indicated in #634 , the correct flags for Intel C++11 on linux is -std=c++11.

I have no idea why the change in history.adoc is here and not in the boostorg repo.

Anyway, the 2 lines in the build.sh script are what matters.


ax3l:
There is a bit more to this, e.g. the [Intel oneAPI ICC](https://software.intel.com/content/www/us/en/develop/articles/oneapi-repo-instructions.html) activation scripts are now in
```
/opt/intel/inteloneapi/setvars.sh
```

src/engine/build.sh
```diff
@@ -81,7 +81,7 @@ test_cxx11 ()
         case $1 in
             gcc) ( ${CXX:=g++} -x c++ -std=c++11 check_cxx11.cpp && rm -f a.out ) 1>/dev/null 2>/dev/null ;;
             intel-darwin) ( ${CXX:=icc} -xc++ check_cxx11.cpp && rm -f a.out ) 1>/dev/null 2>/dev/null ;;
-            intel-linux) ( ${CXX:=icc} -xc++ check_cxx11.cpp && rm -f a.out ) 1>/dev/null 2>/dev/null ;;
+            intel-linux) ( ${CXX:=icc} -std=c++11 check_cxx11.cpp && rm -f a.out ) 1>/dev/null 2>/dev/null ;;
```
Flamefire, line 84: Isn't the meaning of `-xc++` "Treat the source files as C++ (as opposed to C)" and not related to the standard?

aminiussi: >isn't the meaning of -xc++ "Treat the source files as C++ (as opposed to C)" and not related to the standard?  

I don't know, it's not a recognized option of my linux icc.  I would have expected:
```
             intel-linux) ( ${CXX:=icpc} -std=c++11 check_cxx11.cpp && rm -f a.out ) 1>/dev/null 2>/dev/null ;; 
```
The build error is connected to the fact that `icc -xc++` fails.

> But is there an actual failure due to this flag missing?

Yes, as some code (in test for example)  requires c++11 to be activated, which is not the case by default.

aminiussi: > Your branch is outdated, please rebase the branch to remove the confusing part. For example, assuming boostorg is listed in git remote -v as "mainline": git fetch mainline; git rebase mainline/develop.

Thanks, I'll try that

> General question: are these changes sufficient for you to build with a recent ICC on Linux? 

Yes

> Cannot reproduce that this already fixes all the outdated parts when trying this in alpaka-group/alpaka#1066

My Intel distribution is slightly older.

```
[alainm@castor ~]$ icpc -v
icpc version 19.1.0.166 (gcc version 8.3.0 compatibility)
[alainm@castor ~]$ 
```

Flamefire: I've just ran `icc --help` on an icc I could get hold of and it does output `-x <type>`, so only thing missing is the space, although I tried it and it doesn't error out.

Can you elaborate what you mean by "it's not a recognized option of my linux icc"? Is there an actual error when you use it on a test file? Does it really not show up in `icc --help` (it does for me)?

I've also been able to successfully compile a C++11 source without any additional flag. That was `icc 18.0.1 20171018`. But with an older version (`icc 15.0.1 20141023`) it did require the `-std` flag.

Hence I'd suggest to change this line:

```suggestion
            intel-linux) ( ${CXX:=icc} -x c++ -std=c++11 check_cxx11.cpp && rm -f a.out ) 1>/dev/null 2>/dev/null ;;
```

and remove the other change because that is unrelated. I assume the `-std=` flag will be added later if this check here succeeds. And for completeness I think the intel-darwin should be changed too.

aminiussi: > Can you elaborate what you mean by "it's not a recognized option of my linux icc"? Is there an actual error when you use it on a test file? Does it really not show up in icc --help (it does for me)?

The -xc++ option (no space) is not in my documentation. In my manpage  `-x<something>` refers to code generation (where something is  an architecture like **BROADWELL** or **COMMON-AVX512**).

I tried both on a dummy source file and there was no error. The build (bootstrap.sh actually) failed on a c++11 construct that I misattributed apparently.

As for C++11 compatibility, there are two aspects: the library selection, based on the underlying g++ distribution, the dialect selection, which is impacted by the `-std=...` option (default can cover the standard in various way according to distributions I guess). 

I'm not sure which the script insist on using icc instead of icpc which is the normal Intel's C++ compiler name.

Flamefire: Yes because the option is documented as `-x <type>`. `icc` is used instead of `icpc` presumably for mirroring behavior of GCC and `icc` is fully able to compile C++ code. Either by checking the file extension or by `-x c++`.

So I guess if you rebase your branch, fix the `-xc++` to `-x c++' typo and add (not replace) `-std=c++11` then this PR can be accepted.

aminiussi: > icc is used instead of icpc presumably for mirroring behavior of GCC 

But with gcc, g++ is used:
> gcc) ( ${CXX:=g++} -x c++ -std=c++11 check_cxx11.cpp && rm -f a.out ) 1>/dev/null 2>/dev/null ;;

Flamefire: Then it was likely overlooked in https://github.com/boostorg/build/commit/0ec79867fd4605738cfea3333f057ae02dfd7417 which changed gcc to g++ and similar. I guess it is fine to change icc to icpc then

aminiussi:
The rebase has too many conficts. I'll start from a fresh repository

aminiussi:
Follow up in #637

src/engine/build.sh
ax3l: The Intel C++ compiler is technically `icpc`. The C compiler `icc` does recognize C++ files and forwards, but technically we should call `icpc`.

PR #637, opened 2020-07-13
The correct option to activate C++11 with Intel compiler is -std=c++11

aminiussi:
This is a followup of #635


src/engine/build.sh
```diff
@@ -81,7 +81,7 @@ test_cxx11 ()
         case $1 in
             gcc) ( ${CXX:=g++} -x c++ -std=c++11 check_cxx11.cpp && rm -f a.out ) 1>/dev/null 2>/dev/null ;;
             intel-darwin) ( ${CXX:=icc} -xc++ check_cxx11.cpp && rm -f a.out ) 1>/dev/null 2>/dev/null ;;
-            intel-linux) ( ${CXX:=icc} -xc++ check_cxx11.cpp && rm -f a.out ) 1>/dev/null 2>/dev/null ;;
+            intel-linux) ( ${CXX:=icc} -x c++ -std=c++11 check_cxx11.cpp && rm -f a.out ) 1>/dev/null 2>/dev/null ;;
```
ax3l, line 84:
```suggestion
            intel-linux) ( ${CXX:=icpc} -x c++ -std=c++11 check_cxx11.cpp && rm -f a.out ) 1>/dev/null 2>/dev/null ;;
```

ax3l:
As documented [here](https://github.com/boostorg/build/pull/635#issuecomment-657264482), please also update the ancient auto-detect logic to source `/opt/intel/inteloneapi/setvars.sh`

ax3l:
Using the current PR as is I still get build errors of the form
```
icc -o bootstrap/jam0 command.c compile.c constants.c debug.c execcmd.c frames.c function.c glob.c hash.c hdrmacro.c headers.c jam.c jambase.c jamgram.c lists.c make.c make1.c object.c option.c output.c parse.c pathsys.c regexp.c rules.c scan.c search.c subst.c timestamp.c variable.c modules.c strings.c filesys.c builtins.c class.c cwd.c native.c md5.c w32_getreg.c modules/set.c modules/path.c modules/regex.c modules/property-set.c modules/sequence.c modules/order.c execunix.c fileunix.c pathunix.c
./build.sh: 17: ./build.sh: icc: not found
```
when trying to use `intel-linux` on Ubuntu 18.04 when installing ICC from oneAPI: https://software.intel.com/content/www/us/en/develop/articles/oneapi-repo-instructions.html

src/engine/build.sh
aminiussi: Ok, but should we do a query-replace ? some icc seems to be looked up in absolute path that probably made sense on someone's installation.

Or should we replace all of them but the absolute path ?

aminiussi:
> As documented [here](https://github.com/boostorg/build/pull/635#issuecomment-657264482), please also update the ancient auto-detect logic to source `/opt/intel/inteloneapi/setvars.sh`

But that path is specific to an installation. Do we know on what machine host that installation and do we have access to it ?

Flamefire:
I think it makes sense to add that path to the logic here: https://github.com/boostorg/build/blob/b207584f8c99575330121b60d5b844079aaf9663/src/engine/build.sh#L128-L142

They use similar style checks. It is a heuristic and as such IMO fine, although I'd place such stuff at the bottom of the if-elseif chain but it already is like that :shrug:

Flamefire:
Please check how to use `git rebase -i origin/master` after `git fetch origin` (if `origin` is the name of the boostorg/build remote)

aminiussi:
> Please check how to use `git rebase -i origin/master` after `git fetch origin` (if `origin` is the name of the boostorg/build remote)

Should I rebase on master or develop ?

aminiussi:
> > > Please check how to use `git rebase -i origin/master` after `git fetch origin` (if `origin` is the name of the boostorg/build remote)
> > 
> > 
> > Should I rebase on master or develop ?
> 
> develop

Ok, too messy to roll back then, I'll do a new pull request.

aminiussi:
Ok pull request #639  should do it.

PR #639, opened 2020-07-15
Bugfix/634 intel linux icpc

aminiussi:
Follow up to #637

### DIFF
--- a/src/engine/build.sh
+++ b/src/engine/build.sh
@@ -81,7 +81,7 @@ test_cxx11 ()
         case $1 in
             gcc) ( ${CXX:=g++} -x c++ -std=c++11 check_cxx11.cpp && rm -f a.out ) 1>/dev/null 2>/dev/null ;;
             intel-darwin) ( ${CXX:=icc} -xc++ check_cxx11.cpp && rm -f a.out ) 1>/dev/null 2>/dev/null ;;
-            intel-linux) ( ${CXX:=icc} -xc++ check_cxx11.cpp && rm -f a.out ) 1>/dev/null 2>/dev/null ;;
+            intel-linux) ( ${CXX:=icpc} -xc++ check_cxx11.cpp && rm -f a.out ) 1>/dev/null 2>/dev/null ;;
             vacpp) ( ${CXX:=xlC_r} check_cxx11.cpp && rm -f a.out ) 1>/dev/null 2>/dev/null ;;
             xlcpp) ( ${CXX:=xlC_r} check_cxx11.cpp && rm -f a.out ) 1>/dev/null 2>/dev/null ;;
             como) ( ${CXX:=como} check_cxx11.cpp && rm -f a.out ) 1>/dev/null 2>/dev/null ;;
@@ -124,7 +124,7 @@ guess_toolset ()
     elif test_uname FreeBSD && test_path freebsd-version && test_path clang++ && test_cxx11 clang ; then B2_TOOLSET=clang
     elif test_path g++ && test_cxx11 gcc ; then B2_TOOLSET=gcc
     elif test_path clang++ && test_cxx11 clang ; then B2_TOOLSET=clang
-    elif test_path icc && test_cxx11 intel-linux ; then B2_TOOLSET=intel-linux
+    elif test_path icpc && test_cxx11 intel-linux ; then B2_TOOLSET=intel-linux
     elif test -r /opt/intel/cc/9.0/bin/iccvars.sh && test_cxx11 intel-linux ; then
         B2_TOOLSET=intel-linux
         B2_TOOLSET_ROOT=/opt/intel/cc/9.0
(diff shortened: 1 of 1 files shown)


src/engine/build.sh
```diff
@@ -124,7 +124,10 @@ guess_toolset ()
     elif test_uname FreeBSD && test_path freebsd-version && test_path clang++ && test_cxx11 clang ; then B2_TOOLSET=clang
     elif test_path g++ && test_cxx11 gcc ; then B2_TOOLSET=gcc
     elif test_path clang++ && test_cxx11 clang ; then B2_TOOLSET=clang
-    elif test_path icc && test_cxx11 intel-linux ; then B2_TOOLSET=intel-linux
+    elif test_path icpc && test_cxx11 intel-linux ; then B2_TOOLSET=intel-linux
+    elif test -r /opt/intel/inteloneapi/setvars.sh && test_cxx11 intel-linux ; then
+        B2_TOOLSET=intel-linux
+        B2_TOOLSET_ROOT=/opt/intel/inteloneapi
```
ax3l, line 130: Looking at the logic below, I think this should read:
```suggestion
        B2_TOOLSET_ROOT=/opt/intel/inteloneapi/compiler/latest/linux/
```

ax3l: Unfortunately, the `iccvars.sh` script in `bin/` seems to be a dead link on my local copy of oneAPI pulled via the apt repos: https://software.intel.com/content/www/us/en/develop/articles/oneapi-repo-instructions.html

We should instead source `/opt/intel/inteloneapi/setvars.sh` below with this one.

aminiussi: I do not have that distribution right now. So if anyone with that distribution can confirm I'll commit the suggestion.

ax3l: I'll post a suggestion, give me a sec :)

src/engine/build.sh
```diff
@@ -272,7 +275,7 @@ case $B2_TOOLSET in
         fi
         if test -r ${B2_TOOLSET_ROOT}bin/iccvars.sh ; then
```
ax3l, line 276:
```suggestion
        if test -r ${B2_TOOLSET_ROOT}bin/iccvars.sh ; then
            . ${B2_TOOLSET_ROOT}/setvars.sh
        elif test -r ${B2_TOOLSET_ROOT}bin/iccvars.sh ; then
```

src/engine/build.sh
ax3l: @aminiussi can you please mark this one as "resolved"? :)

src/engine/build.sh
ax3l: @aminiussi can we apply this one please?

src/engine/build.sh
```diff
@@ -247,12 +250,12 @@ case $B2_TOOLSET in
     ;;
 
     intel-linux)
-        CXX=${CXX:=icc}
+        CXX=${CXX:=icpc}
         CXX_VERSION_OPT=${CXX_VERSION_OPT:=--version}
         test_path ${CXX} >/dev/null 2>&1
         if test $? ; then
```
ax3l, line 256: Spotted one more bug in the previous implementation:
```suggestion
        if test_path ${CXX} ; then
```

src/engine/build.sh
```diff
@@ -270,9 +273,11 @@ case $B2_TOOLSET in
                 B2_TOOLSET_ROOT=/opt/intel/compiler50/ia32/
             fi
         fi
-        if test -r ${B2_TOOLSET_ROOT}bin/iccvars.sh ; then
+        if test -r ${B2_TOOLSET_ROOT}/setvars.sh ; then
+            . ${B2_TOOLSET_ROOT}/setvars.sh
```
ax3l, line 277: great, the last problem is that this script uses bash-ism...

oh dear, ... not sure if we want to promote the whole suite here (see line 1) to `bash` for that purpose... I will complain upstream with intel, everything else is potentially okay and can already go in.

This will already work if you apply the patch a few lines above:
```
. /opt/intel/inteloneapi/setvars.sh
./bootstrap.sh --with-toolset="intel-linux"
```

aminiussi: Just after:
```
            echo "No intel compiler in current path"
            echo "Look in a few old place for legacy reason"
```
Shouldn't we change the comment then, as oneapi is not that old ?

ax3l: I think the "legacy reasons" are that the script always tried to find the paths, not that the compiler it looks for is old.

aminiussi:
So I think everything is there. Let me know if there is anything missing.

src/engine/build.sh
```diff
@@ -247,18 +250,19 @@ case $B2_TOOLSET in
...
-            B2_TOOLSET_ROOT=`echo ${CXX}| sed -e 's/bin.*\/icc//'`
+            B2_TOOLSET_ROOT=`echo ${CXX}| sed -e 's/bin.*\/icpc//'`
             # probably the most widespread
             ARCH=intel64
         else
             echo "No intel compiler in current path"
-            echo "Look in a few old place for legacy reason"
-            if test -r /opt/intel/cc/9.0/bin/iccvars.sh ; then
+            echo "Look in a few common places just in case"
+	    if test -r /opt/intel/inteloneapi/setvars.sh ; then
```
ax3l, line 263: Just a minor formatting problem:
```suggestion
            if test -r /opt/intel/inteloneapi/setvars.sh ; then
```

ax3l:
@aminiussi thank you! I annotated a last formatting quirk and with the mentioned work-around of activating icc beforehand I can execute this now with ICC from oneAPI in beta07.
There are individual libraries like Boost.Log and Boost.Asio that do not build, but that's not part of this PR enabling icc again in general :)

aminiussi:
> It all looks good.. But I have one final small request.. Can you add appropriate copyright attribution statements?

Sure, can you be more specific ? I don't know the story of that section very much.

PR #642, opened 2020-08-06
use '-arch arm64' for 64-bit builds using darwin.jam

aumuell:
Even for 64-bit architectures, b2 would add '-arch arm' to the compiler
options - but this is only valid for 32-bit arm builds.
The problem was also observed here: https://stackoverflow.com/a/47096479


michaelld:
This change is a good place to start for adding in ARM64 support to Boost. It covers the non-universal build case, which is what most users will need. It does not handle universal installs -- x86_64 + ARM64 -- which will require some added code and logic since Apple is introducing a new architecture & the current darwin.jam logic is not robust enough to directly handle it.

aumuell:
Thank you!

PR #651, opened 2020-09-04
Enable building with clang on Windows

Gei0r:
This adds awareness for building `b2` with `clang++` on Windows.
The flags are the same as for `g++`.

Note that this depends on https://github.com/boostorg/boost/commit/202c67acdd6033c431a2b6ad776bdf480be898d8, which is not yet in boost's master branch (see https://github.com/boostorg/boost/issues/427).


Gei0r:
Why has this change not landed in boost 1.75.0?

PR #660, opened 2020-09-24
Add ASAN support for MSVC

mikekazakov:
Hi,

This change adds Address Sanitizer support for MSVC 16.7+.

The flags try to follow the guidelines from these blogposts:
https://devblogs.microsoft.com/cppblog/addresssanitizer-asan-for-windows-with-msvc/
https://devblogs.microsoft.com/cppblog/asan-for-windows-x64-and-debug-build-support/

Various configurations are tested in this repo via GitHub Actions:
https://github.com/mikekazakov/build-asan-test

The following combinations are tested:
(Regular / ASAN) x (x86 / x64) x (exe / dll / unit-test) x (static runtime / dynamic runtime).

The Jam code doesn't looks very clean due to the requirement of having different flags for "msvc.link" and "msvc.link.dll". Possibly there's a better way of encoding it, but I haven't found it yet.

### DIFF
--- a/src/tools/msvc.jam
+++ b/src/tools/msvc.jam
@@ -485,6 +485,29 @@ rule configure-version-specific ( toolset : version : conditions )
         toolset.flags $(toolset).compile CFLAGS $(conditions) : "/Zc:throwingNew" ;
     }
 
+    # 14.27 (VS2019 Version 16.7) introduced support for ASAN on x86 and x64 CPUs
+    # This check however now only tests for 14.2 (which is 16.0) as msvc.jam doesn't distinguish between minor versions (e.g. 14.21..14.28 etc)
+    if ! [ version.version-less [ SPLIT_BY_CHARACTERS [ MATCH "^([0123456789.]+)" : $(version) ] : . ] : 14 2 ]
+    {
+        # Declare flags for the address sanitizer.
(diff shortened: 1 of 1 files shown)


src/tools/msvc.jam
```diff
@@ -1880,6 +1881,10 @@ local rule register-toolset-really ( )
     toolset.flags msvc.compile INCLUDES <include> ;
     toolset.flags msvc.compile FORCE_INCLUDES <force-include> ;
 
+    # Declare flags for the address sanitizer.
+    toolset.flags msvc.compile.c OPTIONS <address-sanitizer>on : -fsanitize=address /FS ;
+    toolset.flags msvc.compile.c++ OPTIONS <address-sanitizer>on : -fsanitize=address /FS ;
```
grafikrobot, line 1886: Please use the "/" style options to be consistent with the rest of the options use for msvc.

grafikrobot (changes requested):
Seeing as the address sanitizer is only available on recent msvc version it would be better to condition the addition of the options to the versions that support it. Which means adding them in the `configure-version-specific` rule. Can you do those changes?

mikekazakov:
I changed the flag prefix to a slash and moved the feature under a version check.
However, it only tests if the MSVC version is 14.2+ (2019+) as there's no infrastructure to check for minor versions in msvc.jam.
Now it considers all MSVC 2019 16.0-16.8 versions to be 14.2.

PR #669, opened 2020-11-06
Fix memory leak in startup::bootstrap (boostorg#668)

mst7555:
The executable_path allocates a string (strdup) which was never released

Tested with ASAN and valgrind


mst7555:
The problem appeared in 1.74 so it should be backported

PR #670, opened 2020-11-11
Change: Throw error in build.sh executed by bootstrap.

MaxSac:
If an error occures due testing if the compiler is capable to build boost,
it would be nice to locate the error.
Otherwise it is more complicated than it should be to trace the error back.


MaxSac:
These are all valid and good suggestions. I try to submit a first proposal until tomorrow evening.

PR #675, opened 2020-11-25
Ensure Embarcadero toolset name is only "embtc"

tanzislam:
This change, along with boostorg/config#346, fixes the expected library names in a versioned-layout build.

### DIFF
--- a/src/tools/common.jam
+++ b/src/tools/common.jam
@@ -929,6 +929,7 @@ local rule toolset-tag ( name : type ? : property-set )
         case cw : tag += cw ;
         case darwin* : tag += xgcc ;
         case edg* : tag += edg ;
+        case embarcadero* : tag += bcb ;
         case gcc* :
         {
             switch [ $(property-set).get <target-os> ]
--- a/src/tools/common.jam
+++ b/src/tools/common.jam
@@ -929,7 +929,7 @@ local rule toolset-tag ( name : type ? : property-set )
         case cw : tag += cw ;
         case darwin* : tag += xgcc ;
         case edg* : tag += edg ;
-        case embarcadero* : tag += bcb ;
+        case embarcadero* : tag += embtc ;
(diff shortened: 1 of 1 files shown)


eldiener:
The actual toolset for the Embarcadero C++ clang-based compiler, using clang command line parameters, is embarcadero.jam and not borland.jam.

tanzislam:
I can change the name from `bcb` to `embtc`, but that would mean:

 - **either:** enlisting "embtc" [here](https://github.com/boostorg/build/blob/develop/src/tools/common.jam#L996-L1002) to avoid adding the version to the toolset name.
 - **or:** having to maintain per-version toolset names in [`auto_link.hpp`](https://github.com/boostorg/config/blob/develop/include/boost/config/auto_link.hpp#L190-L197), similar to the MSVC cases above.

What do you think?

eldiener:
I think the embarcadero toolset should be different, with 'embtc' in both places. I do not understand the purpose of per-version toolset names.

tanzislam:
>I do not understand the purpose of per-version toolset names.

It is to avoid any binary-incompatibility introduced between compiler releases, similar to the incompatibility between the classic and Clang-based compilers. Ideally users should recompile Boost with the compiler version they are using for their application code (especially when they are upgrading to a newer compiler release), and the "versioned" naming layout tries to protect against users forgetting to do so.

For now, I avoided encoding the release version in the toolset name, as C++Builder seems to be only community-supported, so having to update `auto_link.hpp` with each C++Builder release will become a maintenance chore.

tanzislam:
@eldiener Does this look good to you? Since you're listed as a reviewer, your approval would most likely be needed before the maintainer(s) of the two repositories can decide when to merge these PRs into "develop" and later "master" for release.

eldiener:
Looks fine to me.

PR #676, opened 2020-11-28
Basic changes for GNU/Hurd

pinotree:
This series adds basic support for GNU/Hurd in b2. Luckily there are not many changes needed, it mostly works as it is already.

### DIFF
--- a/src/engine/sysinfo.cpp
+++ b/src/engine/sysinfo.cpp
@@ -18,7 +18,7 @@
 #include <unistd.h>
 #endif
 
-#if defined(OS_LINUX)
+#if defined(OS_LINUX) || defined(__GLIBC__)
 // Need to define this in case it's not as that's the only way to get the
 // sched_* APIs.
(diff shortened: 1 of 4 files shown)


src/engine/sysinfo.cpp
```diff
@@ -18,7 +18,7 @@
 #include <unistd.h>
 #endif
 
-#if defined(OS_LINUX)
+#if defined(OS_LINUX) || defined(__GLIBC__)
```
grafikrobot, line 21: Shouldn't that be `|| defined(OS_HURD)` ?

pinotree: As I explained in the commit message, this applies to anything that uses the GNU libc (glibc).

grafikrobot: Sorry about that.. The way GitHub hided commit messages makes for me missing many of them when reviewing PRs :-(

PR #683, opened 2020-12-13
Adding top-level index.html which redirects to the doc/html/index.htm…

eldiener:
…l, where the documentation gets built by default. A Boost release expects a top-level index.html brings up the documentation for a library or a tool.

### DIFF
--- a/index.html
+++ b/index.html
@@ -1,0 +1,16 @@
+<html>
+<head>
+<meta http-equiv="refresh" content="0; URL=doc/html/index.html">
+</head>
+<body>
+Automatic redirection failed, please go to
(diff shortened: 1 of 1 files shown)


eldiener:
For any given Boost release the documentation link expects an 'index.html' in a library or tool's top-level directory. Open, for example, the Boost 1.75 documentation in your browser as distributed by the 'index.html' of that distribution in the Boost top-level directory. Now click 'Tools' and click the Boost.Build link. You'll get a page not found becuase the Boost documentation expects an 'index.html' in the 'tools/build' subdirectory. My PR adds that file, which redirects to the actual Boost.Buil;d documentation in the distribution.

PR #686, opened 2020-12-17
Cygwin relpath fix

### LINKED ISSUE #684: Broken cygwin b2 start
Cygwin makes executable_path() go under the fallback path:
https://github.com/boostorg/build/blob/develop/src/engine/jam.cpp#L780

If the user calls b2 with a relative path (./b2) it will return NULL string and make the program terminate by uncaught logical_error exception, when it tries to construct a std::string from that executable_path():
https://github.com/boostorg/build/blob/develop/src/engine/startup.cpp#L159-L160

A workaround for a user is to call b2 with a full path:  $PWD/b2   although it is not evident as there is no message about that.

A fix would be to use the same executable_path() implementation as linux, as /proc/self/exe is the same kind of symlink. And maybe startup.cpp should handle the NULL result better.

viric:
It crashed otherwise (executable_path fallback). Should fix #684

### DIFF
--- a/src/engine/jam.cpp
+++ b/src/engine/jam.cpp
@@ -747,7 +747,7 @@ char * executable_path( char const * argv0 )
     sysctl( mib, 4, buf, &size, NULL, 0 );
     return ( !size || size == sizeof( buf ) ) ? NULL : strndup( buf, size );
 }
-#elif defined(__linux__)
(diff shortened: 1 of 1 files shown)


viric:
AH! develop/ is fixed! no need for any patch.

PR #700, opened 2021-01-08
Fix for icl.exe in the Intel OneAPI product

eldiener:
The icl.exe major version number in the oneAPI product starts with "2021" rather than "21", but the fix supports also specifying the major version number of icl.exe in the oneAPI product as starting with "21". In the oneAPI product the setup file is called "setvars.bat" and not "iclvars.bat". Finally all tests for icl major versions are changed from the previously faulty string comparisons to the correct number comparisons.

### DIFF
--- a/src/tools/intel-win.jam
+++ b/src/tools/intel-win.jam
@@ -16,6 +16,7 @@ import toolset ;
 import generators ;
 import type ;
 import path ;
+import numbers ;
 
 feature.extend-subfeature toolset intel : platform : win ;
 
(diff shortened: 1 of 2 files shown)


eldiener:
Please merge.

UpasakhaJason:
On Windows, the OneAPI installation has a different directory structure than the classic Parallel Studio installation.  It looks like intel-win.jam is still assuming the same directory structure for icl version 2021.1.  Is this something that will be addressed?

eldiener:
It should be addresses in this fix. Please take a look at how it is determined where the icl.exe is compared to the root directory.

PR #717, opened 2021-03-26
Handle spaces in CXX path in config_toolset.bat

NeroBurner:
Surround the `CXX` path with double quotes to support spaces in the path.
Otherwise if the `CXX` variable is set to a path containing spaces like

```
C:/Program Files (x86)/Microsoft Visual Studio/2017/BuildTools/VC/Tools/MSVC/14.16.27023/bin/Hostx64/x64/cl.exe
```

the following error message is printed:

```
'C:/Program' is not recognized as an internal or external command,
operable program or batch file.
```


mloskot:
https://lists.boost.org/Archives/boost/2021/03/251031.php says:

> And hence PRs for B2 will need to be posted to the new repo. 

https://github.com/bfgroup/b2

NeroBurner:
opened https://github.com/bfgroup/b2/pull/13

should I close this MR?

mloskot:
Yes, I think so.
/cc @grafikrobot

PR #726, opened 2021-05-24
Declare the /python//numpy target to hold the include path (fixes #725)

### LINKED ISSUE #725: Boost.NumPy doesn't build when more than one Python version is configured in user-config
When user-config.jam has more than one version of Python configured, e.g.
```
using python : 2.7 : C:/Python27 ;
using python : 3.9 : C:/Python39 ;
```
trying to build Python (with e.g. `b2 --with-python python=2.7`) fails, because Boost.NumPy is built against the (correct) 2.7 version of Python, but against the 3.9 headers of NumPy:
```
cl /Zm800 -nologo "libs\python\src\numpy\ufunc.cpp" -Fo"bin.v2\libs\python\build\msvc-14.2\debug\address-model-32\link-static\python-2.7\threading-multi\numpy\ufunc.obj"    -TP /wd4675 /EHs /GR /Zc:throwingNew /Z7 /Od /Ob0 /W3 /MDd /Zc:forScope /Zc:wchar_t /Zc:inline -c -DBOOST_ALL_NO_LIB=1 -DBOOST_NUMPY_SOURCE -DBOOST_NUMPY_STATIC_LIB -DBOOST_PYTHON_STATIC_LIB "-I." "-IC:\Python27\Include" "-IC:\Python39\lib\site-packages\numpy\core\include"
```
The reason for that is that the Python configuration in b2's `python.jam` stores the NumPy include path in a variable `.numpy-include`, which can only hold one value.

https://github.com/boostorg/build/blob/916afd3876bc0d5b706a63f9d09ba66ff69e7d66/src/tools/python.jam#L1087-L1090
https://github.com/boostorg/build/blob/916afd3876bc0d5b706a63f9d09ba66ff69e7d66/src/tools/python.jam#L858

This can in principle be fixed by declaring a target for NumPy's include path, instead of a variable, and then linking to it instead of using `<include>$(numpy-include)` in build/Jamfile.

https://github.com/boostorg/python/blob/2a82afdf6d30c28eb1775f3b8041d988f6f04598/build/Jamfile#L120

pdimov:
This declares a `/python//numpy` target in `python.jam` to hold the NumPy include path; the current practice of using a variable does not support more than one Python installation. See issue #725.

### DIFF
--- a/src/tools/python.jam
+++ b/src/tools/python.jam
@@ -1052,6 +1052,9 @@ local rule configure ( version ? : cmd-or-prefix ? : includes * : libraries ? :
             ;
     }
 
+    # Declare the numpy target, which contains the NumPy include directory
+
+    alias numpy : : $(target-requirements) : : <include>$(.numpy-include) ;
 }
 
 # Conditional rule specification that will prevent building of a target


SoapGentoo:
I've verified that this patch works with multiple python implementations, please merge it.

thesamesam:
I had the same issue but this patch (combined with the Boost Python fix) allows a successful build against the correct numpy headers:


```
"x86_64-pc-linux-gnu-g++"   -fvisibility-inlines-hidden -O2 -pipe -march=native -fdiagnostics-color=always -frecord-gcc-switches -std=c++14 -fPIC -m64 -pthread -finline-functions -Wno-inline -Wall -fvisibility=hidden  -DBOOST_ALL_NO_LIB=1 -DBOOST_NUMPY_SOURCE -DNDEBUG  -I"." -I"/usr/include/python3.9" -I"/usr/lib/python3.9/site-packages/numpy/core/include"  -c -o "bin.v2/libs/python/build/gcc-11.1/gentoorelease/pch-off/python-3.9/threading-multi/visibility-hidden/numpy/ufunc.o" "libs/python/src/numpy/ufunc.cpp"
[...]
"x86_64-pc-linux-gnu-g++"   -fvisibility-inlines-hidden -O2 -pipe -march=native -fdiagnostics-color=always -frecord-gcc-switches -std=c++14 -fPIC -m64 -pthread -finline-functions -Wno-inline -Wall -fvisibility=hidden  -DBOOST_ALL_NO_LIB=1 -DBOOST_NUMPY_SOURCE -DNDEBUG  -I"." -I"/usr/include/python3.8" -I"/usr/lib/python3.8/site-packages/numpy/core/include"  -c -o "bin.v2/libs/python/build/gcc-11.1/gentoorelease/pch-off/python-3.8/threading-multi/visibility-hidden/numpy/ufunc.o" "libs/python/src/numpy/ufunc.cpp"
```

Downstream bug in Gentoo: https://bugs.gentoo.org/733830.

